PR #51, opened 2022-12-02
feat(Hardhat): Scaffold initial Hardhat container

weiyuan95:
- Scaffold the initial implementation of Hardhat Network into a package. This is for testing purposes for integration tests with the server.
- Adds ci/build and ci/test github workflows
- Also added ci/test_hardhat, for the eventual smart contract testing we are doing to do

#### Additional comments?:
- This does not make the server side changes, if not this will be a huge PR. Will be creating another PR with the relevant server side changes
- On test naming convention:
  - `unit` tests are tests that usually involve pure functions, eg. writing a unit test for `addThenSubtract`
  - `i9n` (integration) tests are tests which involve testing the integration of components. Will almost always involve a docker container in our case.
  - `e2e` tests are tests which mimic how an end user would use the application stack. In the case of the bridge, this would range from the front end to the transaction correctly hitting the blockchain.

#### Developer Checklist:

- [x] Unit tests\*


.npmrc
```diff
@@ -9,4 +9,5 @@ public-hoist-pattern[]=*prettier*
 public-hoist-pattern[]=@types/*
 
 engine-strict=true
-auto-install-peers=true
\ No newline at end of file
+auto-install-peers=true
+strict-peer-dependencies=false
```
weiyuan95, line 13: This is to disable the warning that comes with the peer dependency issue with react native 18.x versions

apps/server/package.json
```diff
@@ -7,7 +7,7 @@
     "build": "tsc -b ./tsconfig.build.json",
     "clean": "rm -rf dist",
     "lint": "eslint .",
-    "test": "jest",
+    "test": "jest --passWithNoTests",
```
weiyuan95, line 10: Passing in this flag since there are currently no tests for the backend. Will need to remove once tests are added.

packages/smartcontracts/package.json
```diff
@@ -0,0 +1,85 @@
...
+    ],
+    "testTimeout": 300000,
+    "testPathIgnorePatterns": [
+      "/node_modules/",
+      "<rootDir>/src/test"
+    ]
+  },
+  "dependencies": {},
+  "devDependencies": {
+    "ethers": "~5.7.2",
```
thedoublejay, line 55: Should we consider moving this on the root? I guess we might need to use it on web/server?

thedoublejay: Same with the other dev tools on the deps? eslint-config

packages/smartcontracts/src/containers/EvmContractManager.ts
```diff
@@ -0,0 +1,137 @@
...
+
+/**
+ * Class which encapsulates EVM contract logic. Tightly coupled to a HardhatNetworkContainer.
+ *
+ * @param container A started HardhatNetworkContainer
+ * @param signer An ethers signer attached to deployed contracts by default
+ * @param provider An ethers provider used to query the connected chain
+ */
+export class EvmContractManager {
+  #deployedContracts = new Map<string, DeployedContract<any>>();
```
thedoublejay, line 13: You're the first person I saw using the private modifiers for TS :D

weiyuan95: Necessary since we cannot have `private foo` and `get foo` in the same class, but it allows `#foo` and `get foo` 😬

packages/smartcontracts/package.json
weiyuan95: These need to be declared in the package's package.json since when the Hardhat container is built (via the package's Dockerfile), it's built in the context of the package itself. It cannot 'access' hoisted packages.

I wish there was a better way to do this, but so far this just seems to be a limitation of containerising Hardhat. It was never intended to be used this way 🤣

Abhishekkochar (dismissed):
Had a brief look. LGTM

thedoublejay (approved):
Thanks Mr @weiyuan95

PR #149, opened 2022-12-21
feature(setup): initial setup of DeFiChainModule

thedoublejay:
1. Add `docker-compose` that will start a local playground server
2. Created `defichain` module that will host all logic for DeFiChain
3. Add JellyfishJSON support 
4. Enabled CORS and registered helmet
5. Implemented StatsController, just to call Stats 
6. WhaleApiClientProvider (Similar to legacy-api) to create a whale instance based on network.

Need to figure out how to do testing

Fixes #

#### Developer Checklist:

- [ ] Unit tests\*
- [ ] Added e2e tests\*
- [ ] Added translations\*


apps/server/docker-compose.yml
```diff
@@ -0,0 +1,82 @@
+version: '3.7'
+
+services:
+  traefik:
```
weiyuan95, line 4: Just curious, what's the idea of using `traefik` here? To simplify the networking for the exposed services?

apps/server/src/defichain/pipes/NetworkValidationPipe.ts
```diff
@@ -0,0 +1,33 @@
+import { HttpException, Injectable, PipeTransform } from '@nestjs/common';
+
+export type SupportedNetwork = 'mainnet' | 'testnet' | 'regtest' | 'local';
+
+@Injectable()
+export class NetworkValidationPipe implements PipeTransform {
```
weiyuan95, line 6: https://docs.nestjs.com/techniques/validation

Using nestjs' built-in support for `class-validator` is a better end goal to me IMO, since
- All requests objects/interfaces are declared as DTOs, making each request very clear to devs
- Able to leverage on class-validator's robust validation logic decorators

We can implement this down the line since this is just a scaffolding PR, but if we want to use it we should probably adopt it soon before we start writing the different endpoints.

apps/server/src/defichain/controllers/StatsController.ts
```diff
@@ -0,0 +1,15 @@
+import { StatsData } from '@defichain/whale-api-client/dist/api/stats';
```
weiyuan95, line 1:
```suggestion
import { stats } from '@defichain/whale-api-client';
```

And then use `stats.StatsData` below. To prevent dipping into `dist` 😛

apps/server/src/BridgeServerApp.ts
```diff
@@ -21,17 +23,33 @@ export class BridgeServerApp<App extends INestApplication = INestApplication> {
   }
 
   get fastifyAdapter(): FastifyAdapter {
-    return new FastifyAdapter();
+    const adapter = new FastifyAdapter();
+    this.setupJellyfishJSON(adapter);
```
weiyuan95, line 27:
```suggestion
```

This can probably be in `configureApp` as a configure step? ie. 

```ts
  async configureApp(app: NestFastifyApplication): Promise<void> {
    app.useLogger(app.get(Logger));
    app.enableCors({
      origin: '*',
      methods: ['GET', 'PUT', 'POST', 'DELETE'],
      allowedHeaders: ['Content-Type'],
      maxAge: 60 * 24 * 7,
    });
    await app.register(helmet);

    this.setupJellyfishJSON(this.fastifyAdapter);
  }
```

weiyuan95 (commented):
Generally LGTM, just some comments.

Also, IMO it would be good to have some `i9n` tests by spinning up the service and hitting the `StatsController` (to ensure everything works), but I'm guessing this PR is prioritising speed to get things set up.

apps/server/docker-compose.yml
thedoublejay: Yes, same as what we do on LW. We use `19553` to direct traffic for blockchain, whale and playground

apps/server/src/defichain/controllers/StatsController.ts
thedoublejay: damn auto import

apps/server/src/defichain/pipes/NetworkValidationPipe.ts
thedoublejay: Will check this!

thedoublejay:
> Generally LGTM, just some comments.
> 
> Also, IMO it would be good to have some `i9n` tests by spinning up the service and hitting the `StatsController` (to ensure everything works), but I'm guessing this PR is prioritising speed to get things set up.

I'm working today to add a test for StatsController. After that should be ok haha

apps/server/src/BridgeServerApp.ts
thedoublejay: Oh actually this needs to be called prior to creating the app. So it can't be inside the configureApp

weiyuan95 (approved):
Generally LGTM!

PR #172, opened 2023-01-05
feat(SmartContract): daily allowance start time for supported token

### LINKED ISSUE #158: Bridge contract should reset allowance on a daily basis
Each token will reset daily upon a certain time in day.
To do this, when adding a supported token, admin needs to specify a timestamp in the future so that the contract will start supporting the token from that timestamp and reset every day since that timestamp. (e.g. if contract receives 9 Jan 11 pm GMT 00:00.00 as a variable then the contract will start supporting the token from that point in time and will reset every 11 pm GMT 00:00.00)
The admin should also be able to reset this timestamp in the future for each token too.

The implementation for this can follow the below. Copied from [here](https://github.com/WavesHQ/bridge/pull/175#discussion_r1068976178)

> I would suggest that we merge `changeDailyAllowance` and `changeResetAllowanceTime` to one function, `changeTokenAllowance(address _tokenAddress, uint256 _newResetTimestamp, uint256 _dailyAllowance)`. 
>
> With this, the `_newResetTimestamp` essentially controls when the token is active again, and will reject with an appropriate error if `block.timestamp < _newResetTimestamp`.
>
> This simplifies the logic, and we would not need an `inChangeAllowancePeriod` modifier anymore since it's fully controlled from `_newResetTimestamp`.

Keep in mind that the `_newResetTimestamp` _has to be at least 24 hours_ from the original `latestResetTime` of the token so that we do not accidentally 'merge' two allowances for a day.

Abhishekkochar:
Introduces the dailyAllowance start time for supported tokens.
Fixes #158 

#### Additional comments?:
Not certain on the use case yet. Will update accordingly.
#### Developer Checklist:

- [ ] Unit tests\*
- [ ] Added e2e tests\*
- [ ] Added translations\*


packages/smartcontracts/src/tests/BridgeProxy.newSupportedAllowance.spec.ts
```diff
@@ -0,0 +1,28 @@
+import { loadFixture, time } from '@nomicfoundation/hardhat-network-helpers';
```
Abhishekkochar, line 1: @weiyuan95, test for new `dailyAllowance` start time

Abhishekkochar:
Closing this branch. Updated issue and current number of conflicts.

PR #175, opened 2023-01-10
feat(SmartContract): reset daily allowance time for supported token 

### LINKED ISSUE #158: Bridge contract should reset allowance on a daily basis

### LINKED ISSUE #179: EVM -> DFI Smart contract test improvements
When calling `bridgeToDeFiChain`, additional expectations should be added to check that the bridge has actually received the bridged funds.

For example, if the user wants to bridge 10 wBTC, the bridge should have an additional 10wBTC in its balance after the method successfully executes.

#### Why is this needed:
Robust testing!

### LINKED ISSUE #194: `Contracts`: revert if _amount or value is 0
When calling `bridgeToDeFiChain`, users shouldn't allow to send 0 erc20 and ether.

Abhishekkochar:
As the title suggests, Admins can set the `resetDailyAllowanceTime` and  reset with new `resetDailyAllowanceTime`.

Fixes #158
Fixes #179 
Fixes #194 

#### Developer Checklist:

- [ ] Unit tests\*
- [ ] Added e2e tests\*
- [ ] Added translations\*


packages/smartcontracts/src/contracts/BridgeV1.sol
```diff
@@ -271,6 +273,9 @@ contract BridgeV1 is UUPSUpgradeable, EIP712Upgradeable, AccessControlUpgradeabl
         if (_tokenAddress != address(0) && msg.value > 0) {
             revert DO_NOT_SEND_ETHER_WITH_ERC20();
         }
+        uint256 tokenAllowancesStartTime = tokenAllowances[_tokenAddress].prevEpoch;
+        if (tokenAllowancesStartTime > block.timestamp)
```
Abhishekkochar, line 277: @cuongquangnam , please have a quick look when you have time. Thanks

packages/smartcontracts/src/contracts/BridgeV1.sol
```diff
@@ -384,6 +408,20 @@ contract BridgeV1 is UUPSUpgradeable, EIP712Upgradeable, AccessControlUpgradeabl
         emit TRANSACTION_FEE_CHANGED(oldTxFee, transactionFee);
     }
 
+    /**
+     * @notice Use by addresses with Admin and Operational roles to set the new reset allowance for the token
+     * @param _tokenAddress The token supporting new time stamp
+     * @param _newResetEpoch new time stamp in seconds
+     */
+    function changeResetAllowanceTime(address _tokenAddress, uint256 _newResetEpoch) external {
```
cuongquangnam, line 416: Have you thought of the scenario where changeDailyAllowance happens around the same time as changeResetAllowanceTime? 
```typescript
// currently, for tokenA, smart contract resets every 8am 
// at 9am today, the admin changes the daily allowance
await proxyBridge.changeDailyAllowance(tokenA.address, 100);
// later at 10am today, the admin change resetAllowanceTime to 10am tomorrow,
// which makes the resetEpoch = 10 am tomorrow
await proxyBridge.changeResetAllowanceTime(tokenA.address, 10AM_TOMORROW)
// now whenever the modifier notInChangePeriod(tokenA.address) runs, 
// it will check block.timestamp - 10AM_TOMORROW < 1 days 
// instead of check block.timestamp - 8AM_TODAY < 1 days
```

cuongquangnam: I think we should design the contract in such a way that pause periods for changeDailyAllowance and changeResetAllowanceTime do not clash

cuongquangnam (changes requested):

packages/smartcontracts/src/contracts/BridgeV1.sol
cuongquangnam: maybe can take a quick look at some WIP in this branch I did two days ago: https://github.com/WavesHQ/bridge/tree/cuong/reset-passing-timestamp

Abhishekkochar: `_newResetEpoch` & `notInChangeAllowancePeriod == true` can not be in same time frame. Is this what you are referring to?

cuongquangnam: you mean inChangeAllowancePeriod == true? 
Anw, yes, I think smart contract should check for that

Abhishekkochar: Yes, I agree. We should have this check. Are you adding it as part of your PR?

packages/smartcontracts/src/contracts/BridgeV1.sol
```diff
@@ -68,9 +69,14 @@ error TRANSCATION_FAILED();
...
 
+/** @notice @dev 
+/* This error occurs when `_newResetEpoch` is passed block.timestamp
+*/
+error INVALID_RESET_EPOCH_TIME();
+
 contract BridgeV1 is UUPSUpgradeable, EIP712Upgradeable, AccessControlUpgradeable {
     struct TokenAllowance {
-        uint256 prevEpoch;
+        uint256 resetEpoch;
```
weiyuan95, line 79: Wondering if this can be re-named for clarity, since this is more of a `timestamp` and [is definitely not an `epoch`](https://info.etherscan.com/epoch-in-ethereum/#:~:text=In%20Ethereum%2C%20an%20epoch%20is,6.4%20minutes%20in%20total%20time.) 

I'm thinking `lastResetAt`, `lastResetAtTimestamp`, or `latestResetTimestamp`. WDYT?

packages/smartcontracts/src/contracts/BridgeV1.sol
```diff
@@ -384,6 +416,22 @@ contract BridgeV1 is UUPSUpgradeable, EIP712Upgradeable, AccessControlUpgradeabl
         emit TRANSACTION_FEE_CHANGED(oldTxFee, transactionFee);
     }
 
+    /**
+     * @notice Use by addresses with Admin and Operational roles to set the new reset allowance for the token
+     * @param _tokenAddress The token supporting new time stamp
+     * @param _newResetTimeStamp new time stamp in seconds
+     */
+    function changeResetAllowanceTime(address _tokenAddress, uint256 _newResetTimeStamp) external {
```
weiyuan95, line 424: Looking at this, my main worry is that this can potentially cause some unexpected behaviour, and could introduce potential bugs since we have two ways of modifying the state of the `tokenAllowance`. This puts the stress on the devs to ensure that we've covered all the possible states and account for them so that we don't introduce potential attack vectors.

I would suggest that we merge `changeDailyAllowance` and `changeResetAllowanceTime` to one function, `changeTokenAllowance(address _tokenAddress, uint256 _newResetTimestamp, uint256 _dailyAllowance)`. 

With this, the `_newResetTimestamp` essentially controls when the token is active again, and will reject with an appropriate error if `block.timestamp < _newResetTimestamp`.

This simplifies the logic, and we would not need an `inChangeAllowancePeriod` modifier anymore since it's fully controlled from `_newResetTimestamp`?

weiyuan95 (changes requested):

packages/smartcontracts/src/contracts/BridgeV1.sol
Abhishekkochar: As discussed on the call. I'm working on changing the requirements. This might take longer as tests need to be change. Also,  bridging need to tested without  inChangeAllowancePeriod`.

packages/smartcontracts/src/contracts/BridgeV1.sol
```diff
@@ -271,15 +285,18 @@ contract BridgeV1 is UUPSUpgradeable, EIP712Upgradeable, AccessControlUpgradeabl
         if (_tokenAddress != address(0) && msg.value > 0) {
             revert DO_NOT_SEND_ETHER_WITH_ERC20();
         }
+        uint256 tokenAllowancesStartTime = tokenAllowances[_tokenAddress].resetEpoch;
+        if (tokenAllowancesStartTime > block.timestamp)
```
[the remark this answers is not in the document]
Abhishekkochar: Most of them are replacing the traditional require statement for us, resulting in less opcodes.

packages/smartcontracts/src/contracts/BridgeV1.sol
```diff
@@ -266,20 +282,23 @@ contract BridgeV1 is UUPSUpgradeable, EIP712Upgradeable, AccessControlUpgradeabl
...
         if (!supportedTokens[_tokenAddress]) revert TOKEN_NOT_SUPPORTED();
         if (_tokenAddress != address(0) && msg.value > 0) {
             revert DO_NOT_SEND_ETHER_WITH_ERC20();
         }
+        uint256 tokenAllowancesStartTime = tokenAllowances[_tokenAddress].latestResetTimestamp;
+        if (block.timestamp < tokenAllowancesStartTime) revert STILL_IN_CHANGE_ALLOWANCE_PERIOD();
         uint256 amount = checkValue(_tokenAddress, _amount, msg.value);
-        if (tokenAllowances[_tokenAddress].prevEpoch + (1 days) > block.timestamp) {
+        require(amount > 0);
+        if (tokenAllowances[_tokenAddress].latestResetTimestamp + (1 days) > block.timestamp) {
```
weiyuan95, line 294:
```suggestion
        require(amount > 0);
        // Transaction is within the last tracked day's daily allowance
        if (tokenAllowances[_tokenAddress].latestResetTimestamp + (1 days) > block.timestamp) {
```

packages/smartcontracts/src/contracts/BridgeV1.sol
```diff
@@ -297,12 +316,18 @@ contract BridgeV1 is UUPSUpgradeable, EIP712Upgradeable, AccessControlUpgradeabl
...
+    ) external {
         if (!checkRoles()) revert NON_AUTHORIZED_ADDRESS();
         if (supportedTokens[_tokenAddress]) revert TOKEN_ALREADY_SUPPORTED();
+        // Token will be added on the supported list regardless of `_startAllowanceTimeFrom`
         supportedTokens[_tokenAddress] = true;
-        tokenAllowances[_tokenAddress].prevEpoch = block.timestamp;
+        tokenAllowances[_tokenAddress].latestResetTimestamp = _startAllowanceTimeFrom;
         tokenAllowances[_tokenAddress].dailyAllowance = _dailyAllowance;
         tokenAllowances[_tokenAddress].currentDailyUsage = 0;
         tokenAllowances[_tokenAddress].inChangeAllowancePeriod = false;
```
weiyuan95, line 333: We don't need this as well since we don't need to track it anymore.

packages/smartcontracts/src/tests/BridgeProxy.dailyAllowance.spec.ts
```diff
@@ -105,38 +107,41 @@ describe('Daily allowance tests', () => {
...
+      // Increasing time by 1 day
+      await time.increase(currentTimeStamp(60 * 60 * 24));
+      // Bridging 10 test tokens
+      await proxyBridge.bridgeToDeFiChain(ethers.constants.AddressZero, testToken.address, toWei('10'));
+      // No more bridging should be allowed
+      await expect(
+        proxyBridge.bridgeToDeFiChain(ethers.constants.AddressZero, testToken.address, toWei('10')),
+      ).to.revertedWithCustomError(proxyBridge, 'EXCEEDS_DAILY_ALLOWANCE');
+      // Increasing time by 1 day
       await time.increase(60 * 60 * 24);
```
weiyuan95, line 129: Can I verify the expected behaviour here?
1. The token is supposed to be active 2 days in the future.
2. Trying to bridge a transaction causes a revert with STILL_IN_CHANGE_ALLOWANCE_PERIOD
3. 1 day has passed, and `bridgeToDeFiChain` is called again.
4. The revert is now `EXCEEDS_DAILY_ALLOWANCE`

Shouldn't it be `STILL_IN_CHANGE_ALLOWANCE_PERIOD`? 🤔 Since `block.timestamp < tokenAllowanceStartTime` should be `true`

If this test was passing unexpectedly, let's figure out why as well.

packages/smartcontracts/src/tests/BridgeProxy.dailyAllowance.spec.ts
```diff
@@ -105,38 +107,41 @@ describe('Daily allowance tests', () => {
...
+      expect(await testToken.balanceOf(proxyBridge.address)).to.equal(toWei('35'));
     });
     describe('Emitted Events', () => {
       it('Successfully emitted event when changing allowances', async () => {
         const { proxyBridge, testToken, defaultAdminSigner } = await loadFixture(deployContracts);
         await initMintAndSupport(proxyBridge, testToken, defaultAdminSigner.address, proxyBridge.address);
+        const prevTimeStamp = (await proxyBridge.tokenAllowances(testToken.address)).latestResetTimestamp;
         // Event called CHANGE_DAILY_ALLOWANCE should be emitted when changes token's allowances
-        await expect(proxyBridge.changeDailyAllowance(testToken.address, toWei('10')))
+        await expect(proxyBridge.changeDailyAllowance(testToken.address, toWei('10'), currentTimeStamp(60 * 60 * 25)))
```
weiyuan95, line 142: I would set `const currentTime = currentTimeStamp(60 * 60 * 25)`, then use the variable `currentTime` in the expectations.

You never know when the test might suddenly fail because the timestamps are 1ms off 🤣 . This applies to any kind of expectations like this as well where we are expecting 2 timestamps to be the same.

packages/smartcontracts/src/tests/BridgeProxy.evmToDefi.spec.ts
```diff
@@ -13,11 +13,16 @@ async function initMintAndSupport(
   testToken: TestToken,
   eoaAddress: string,
   contractAddress: string,
+  timeStamp?: number,
```
weiyuan95, line 16:
```suggestion
  additionalTime?: number,
```

packages/smartcontracts/src/tests/BridgeProxy.evmToDefi.spec.ts
```diff
@@ -66,76 +71,86 @@ describe('EVM --> DeFiChain', () => {
       ).to.revertedWithCustomError(proxyBridge, 'DO_NOT_SEND_ETHER_WITH_ERC20');
     });
 
-    it('Successfully bridging after a day', async () => {
+    it('Successfully revert if sending zero ERC20 token', async () => {
       const { proxyBridge, testToken, defaultAdminSigner } = await loadFixture(deployContracts);
       await initMintAndSupport(proxyBridge, testToken, defaultAdminSigner.address, proxyBridge.address);
+      // This txn should fail. User sending 0 ERC20 along with ETHER. only checking the _amount not value
+      await expect(
+        proxyBridge.bridgeToDeFiChain(ethers.constants.AddressZero, testToken.address, 0, { value: toWei('10') }),
```
weiyuan95, line 79: If `value: ` here does not matter, let's remove it to remove any possible confusion?

packages/smartcontracts/src/tests/BridgeProxy.evmToDefi.spec.ts
```diff
@@ -66,76 +71,86 @@ describe('EVM --> DeFiChain', () => {
...
-      ).to.revertedWithCustomError(proxyBridge, 'EXCEEDS_DAILY_ALLOWANCE');
-      // Current daily usage should be 12
-      expect((await proxyBridge.tokenAllowances(testToken.address)).currentDailyUsage).to.equal(toWei('12'));
-      // Bridging 3 token again. Txn should not revert.
-      await proxyBridge.bridgeToDeFiChain(ethers.constants.AddressZero, testToken.address, toWei('3'));
-      // Current daily usage should be 15
-      expect((await proxyBridge.tokenAllowances(testToken.address)).currentDailyUsage).to.equal(toWei('15'));
+        proxyBridge.bridgeToDeFiChain(ethers.constants.AddressZero, testToken.address, toWei('10')),
+      ).to.revertedWithCustomError(proxyBridge, 'STILL_IN_CHANGE_ALLOWANCE_PERIOD');
+      // Contract address should be Zero
```
weiyuan95, line 97:
```suggestion
      // Contract address balance should be zero
```

weiyuan95 (changes requested):
Some minor comments. If changing `modifier inChangeAllowancePeriod` causes a lot of bloat to this PR, let's leave it for another one.

packages/smartcontracts/src/contracts/BridgeV1.sol
Abhishekkochar: I'm not sure what state var you talking about.

weiyuan95: Ah sorry, this comment is confusing because it's not a change of this PR.

 I mean we don't need `inChangeAllowancePeriod` anymore since we don't use it anywhere anmore

packages/smartcontracts/src/contracts/BridgeV1.sol
```diff
@@ -150,8 +144,15 @@ contract BridgeV1 is UUPSUpgradeable, EIP712Upgradeable, AccessControlUpgradeabl
      * @notice Emitted when the dailyAllowance of an existing supported token is changed by only Admin accounts
      * @param supportedToken Address of the token being added to supported token
      * @param changeDailyAllowance The new daily allowance of the supported token
+     * @param previousTimeStamp The old timeStamp of the supported toke
+     * @param newTimeStamp The new timeStamp of the supported toke
```
weiyuan95, line 148:
```suggestion
     * @param previousTimeStamp The old reset timeStamp of the supported token that is being replaced
     * @param newTimeStamp The new reset timeStamp of when the supported token starts to be supported
```

packages/smartcontracts/src/tests/BridgeProxy.dailyAllowance.spec.ts
```diff
@@ -101,42 +103,47 @@ describe('Daily allowance tests', () => {
...
+      await proxyBridge.bridgeToDeFiChain(ethers.constants.AddressZero, testToken.address, toWei('10'));
+      // This tx should fail as the dailyAllowance has been met
+      await expect(
+        proxyBridge.bridgeToDeFiChain(ethers.constants.AddressZero, testToken.address, toWei('9')),
+      ).to.revertedWithCustomError(proxyBridge, 'EXCEEDS_DAILY_ALLOWANCE');
+      // Contract balance should be 20 test tokens
+      expect(await testToken.balanceOf(proxyBridge.address)).to.equal(toWei('20'));
     });
     describe('Emitted Events', () => {
       it('Successfully emitted event when changing allowances', async () => {
```
weiyuan95, line 138: We should test the failure scenarios when changing allowances as well for the reset time, I don't see it being tested anywhere
```
        if (_newResetTimeStamp - block.timestamp <= 1 days) revert INVALID_RESET_EPOCH_TIME();
```

packages/smartcontracts/src/tests/BridgeProxy.evmToDefi.spec.ts
```diff
@@ -66,96 +71,98 @@ describe('EVM --> DeFiChain', () => {
...
+        await initMintAndSupport(proxyBridge, testToken, defaultAdminSigner.address, proxyBridge.address, 60 * 60 * 24);
+        // increasing time by 1 day.
         await time.increase(60 * 60 * 24);
-        // Event called BRIDGE_TO_DEFI_CHAIN should be emitted when Successfully bridged token to DefiChain
-        // Getting timestamp
+        await proxyBridge.bridgeToDeFiChain(ethers.constants.AddressZero, testToken.address, toWei('8'));
+        await proxyBridge.bridgeToDeFiChain(ethers.constants.AddressZero, testToken.address, toWei('7'));
+        // Checking Contract balance - should be 15 Test token
+        expect(await testToken.balanceOf(proxyBridge.address)).to.equal(toWei('15'));
+        // Daily allowance set to 10. This tx should fail
```
weiyuan95, line 114:
```suggestion
        // Daily allowance set to 15 during setup. This tx should fail (8 + 7 + 2) > 15
```

packages/smartcontracts/src/tests/BridgeProxy.evmToDefi.spec.ts
```diff
@@ -184,11 +191,22 @@ describe('EVM --> DeFiChain', () => {
...
+
+    it('Successfully revert if sending zero ETHER', async () => {
+      const { proxyBridge, testToken, defaultAdminSigner } = await loadFixture(deployContracts);
+      await initMintAndSupport(proxyBridge, testToken, defaultAdminSigner.address, proxyBridge.address);
+      // This txn should fail. User sending ETHER, will check if the value is greater than 0
+      await expect(
+        proxyBridge.bridgeToDeFiChain(ethers.constants.AddressZero, ethers.constants.AddressZero, 0, {
+          value: 0,
+        }),
+      ).to.reverted;
```
weiyuan95, line 203: According to [the docs](https://hardhat.org/hardhat-chai-matchers/docs/overview#reverts), this should be `to.be.reverted`, let's double check this everywhere!

packages/smartcontracts/src/tests/testUtils/mathUtils.ts
```diff
@@ -10,3 +10,12 @@ export function amountAfterFee(amount: BigNumber, transactionFee: BigNumber): Bi
   const netAmountAfterFee = amount.sub(feeAmount);
   return netAmountAfterFee;
 }
+
+// Current time stamp
+export function currentTimeStamp(additionalTime?: number): number {
```
weiyuan95, line 15: tbh, this should look more like
```ts
function getCurrentTimeStamp({ additionalTime } : GetTimestampOptions): number {...}

interface GetTimestampOptions {
  additionalTime?: number
}
```

So that we can make a cleaner function call like 
```ts
getCurrentTimeStamp({additionalTime: 60 * 60 * 24})
```

but that is a minor thing that belongs to another PR

weiyuan95 (changes requested):
Few more minor changes, then we're good to go!

packages/smartcontracts/src/tests/BridgeProxy.evmToDefi.spec.ts
Abhishekkochar: `to.reverted` and `to.be.reverted` can be used interchangeably. I will change it to to.be.reverted`

packages/smartcontracts/src/tests/testUtils/mathUtils.ts
Abhishekkochar: Okay, I will change this on another PR.

packages/smartcontracts/src/tests/BridgeProxy.dailyAllowance.spec.ts
Abhishekkochar: Done

packages/smartcontracts/src/tests/BridgeProxy.evmToDefi.spec.ts
weiyuan95: Ah, really? Their docs aren't up to date then 🤣

Abhishekkochar: Yes, As per my research. There isn't much difference.

weiyuan95 (approved):

PR #178, opened 2023-01-10
feat(SmartContract): Add tests for Proxy Behaviour

### LINKED ISSUE #162: OpenZeppelin versions should be pinned.
Relevant comment: https://github.com/WavesHQ/bridge/pull/148#discussion_r1060231291

#### Why is this needed:
- We don't introduce any potential breaking changes
- Tests are run one a specific version of the contracts, and these contract versions shouldnt be bumped

cuongquangnam:
Add tests for Proxy behaviour (upgrade + receive Ether)

Fixes #162

#### Developer Checklist:

- [ ] Unit tests\*
- [ ] Added e2e tests\*
- [ ] Added translations\*


packages/smartcontracts/src/tests/BridgeProxy.proxyBehaviour.spec.ts
```diff
@@ -0,0 +1,156 @@
...
+      const { testToken, proxyBridge, defaultAdminSigner } = await loadFixture(deployContracts);
+      blockNum = await ethers.provider.getBlockNumber();
+      console.log('after: ', blockNum);
+      const accounts = await ethers.provider.listAccounts();
+
+      const NewImplementation = await ethers.getContractFactory('NewImplementation');
+      const newImplementation = await NewImplementation.deploy();
+      await newImplementation.deployed();
+
+      let tx = await testToken.mint(proxyBridge.address, 100);
```
Abhishekkochar, line 23: We should remove the `let tx`. Keep this consistent through all tests

packages/smartcontracts/src/tests/BridgeProxy.proxyBehaviour.spec.ts
```diff
@@ -0,0 +1,156 @@
...
+        deadline: ethers.constants.MaxUint256,
+        tokenAddress: testToken.address,
+      };
+
+      const signature = await defaultAdminSigner._signTypedData(domainData, eip712Types, eip712Data);
+      tx = await proxyBridge.claimFund(accounts[2], 10, 0, ethers.constants.MaxUint256, testToken.address, signature);
+      await tx.wait();
+
+      // upgrade to the new contract
+      tx = await proxyBridge.upgradeTo(newImplementation.address);
```
Abhishekkochar, line 55: As mentioned above.

packages/smartcontracts/src/tests/BridgeProxy.proxyBehaviour.spec.ts
```diff
@@ -0,0 +1,156 @@
...
+      proxyBridge = bridgeDeploymentResult.proxyBridge;
+      defaultAdminSigner = bridgeDeploymentResult.defaultAdminSigner;
+    });
+
+    it('Send no data and send some ether to the smart contract', async () => {
+      const prevETHBalance = await ethers.provider.getBalance(proxyBridge.address);
+      const tx = await defaultAdminSigner.sendTransaction({
+        to: proxyBridge.address,
+        data: '0x',
+        value: ethers.utils.parseEther('1.0'),
```
Abhishekkochar, line 98: Please use toWei()

packages/smartcontracts/src/tests/BridgeProxy.proxyBehaviour.spec.ts
```diff
@@ -0,0 +1,156 @@
...
+          data:
+            ethers.utils.keccak256(ethers.utils.toUtf8Bytes('addSupportedTokens(address,uint256)')).substring(0, 10) +
+            abiCoder.encode(['address', 'uint256'], [ethers.constants.AddressZero, 10]).substring(2),
+          value: 100,
+        }),
+      ).to.reverted;
+    });
+
+    it('Call a method that is in the implementation interface that is payable, and send ether', async () => {
+      const tx = await proxyBridge.addSupportedTokens(ethers.constants.AddressZero, 15);
```
Abhishekkochar, line 141: `15` should be `toWei('15')`

Abhishekkochar (changes requested):

packages/smartcontracts/src/contracts/NewImplementation.sol
```diff
@@ -0,0 +1,22 @@
+import '@openzeppelin/contracts-upgradeable/access/AccessControlUpgradeable.sol';
+import '@openzeppelin/contracts-upgradeable/proxy/utils/UUPSUpgradeable.sol';
+import '@openzeppelin/contracts-upgradeable/utils/cryptography/EIP712Upgradeable.sol';
+
+// Inheritance chain is the same as the old version so that the state variables storage slots follow the same order
```
weiyuan95, line 5: Is this necessary and something that we need to ensure for upgraded contracts in the future? What do we need to take note of so that we do not run into problems with the storage slots?

packages/smartcontracts/src/contracts/NewImplementation.sol
```diff
@@ -0,0 +1,22 @@
...
+contract NewImplementation is UUPSUpgradeable, EIP712Upgradeable, AccessControlUpgradeable {
+    // respect the old storage slots order
+    struct TokenAllowance {
+        uint256 prev_epoch;
+        uint256 daily_allowance;
+        uint256 curr_daily_usage;
+        bool in_change_allowance_period;
+    }
+
+    mapping(address => uint256) public eoaAddressToNonce;
```
weiyuan95, line 15: Do these need to be declared in this exact order as well? 😅

packages/smartcontracts/src/tests/BridgeProxy.proxyBehaviour.spec.ts
```diff
@@ -0,0 +1,156 @@
...
+import { ethers } from 'hardhat';
+
+import { BridgeV1 } from '../generated';
+import { BridgeDeploymentResult, deployContracts } from './testUtils/deployment';
+
+describe('Test Behaviour related to proxy', () => {
+  describe('Test Proxy Upgradeability', () => {
+    it('The smart contract retains the values of state variables after upgrading', async () => {
+      let blockNum: Number = await ethers.provider.getBlockNumber();
+      console.log('before ', blockNum);
```
weiyuan95: Let's remove these logs. IMO we don't need to test that the contracts have been successfully deployed since your expectations below will fail if they aren't deployed.

If it's for something else, then let's write a proper expectation like `expect(beforeBlockNumber.lt(afterBlockNumber)).toBeTrue()`

packages/smartcontracts/src/tests/BridgeProxy.proxyBehaviour.spec.ts
weiyuan95: Yep, let's be more descriptive with these instead of reusing `tx` everywhere. This can be `mintTokenTx` for example

weiyuan95 (changes requested):

packages/smartcontracts/src/tests/BridgeProxy.proxyBehaviour.spec.ts
```diff
@@ -0,0 +1,151 @@
...
+      proxyBridge.claimFund(accounts[2], toWei('10'), 0, ethers.constants.MaxUint256, testToken.address, signature);
+
+      // upgrade to the new contract
+      await proxyBridge.upgradeTo(newImplementation.address);
+      // check the implementation slot
+      expect(
+        `0x${(
+          await ethers.provider.getStorageAt(
+            proxyBridge.address,
+            '0x360894a13ba1a3210667c828492db98dca3e2076cc3735a920a3ca505d382bbc',
```
weiyuan95, line 56: Where does this come from? 🤔

packages/smartcontracts/src/tests/BridgeProxy.proxyBehaviour.spec.ts
```diff
@@ -0,0 +1,151 @@
...
+      expect(receipt.logs[0].topics[0]).to.equal(
+        ethers.utils.keccak256(ethers.utils.toUtf8Bytes('ETH_RECEIVED(address,uint256)')),
+      );
+      expect(receipt.logs[0].topics[1]).to.equal(abiCoder.encode(['address'], [defaultAdminSigner.address]));
+      expect(receipt.logs[0].topics[2]).to.equal(abiCoder.encode(['uint256'], [toWei('1')]));
+      const afterETHBalance = await ethers.provider.getBalance(proxyBridge.address);
+      expect(afterETHBalance.sub(prevETHBalance)).to.equal(toWei('1'));
+    });
+
+    it('Call a method that is not in the implementation interface', async () => {
```
weiyuan95, line 107:
```suggestion
    it('should revert when calling a method that is not in the implementation interface', async () => {
```

packages/smartcontracts/src/tests/BridgeProxy.proxyBehaviour.spec.ts
```diff
@@ -0,0 +1,151 @@
...
+    let defaultAdminSigner: SignerWithAddress;
+
+    beforeEach(async () => {
+      abiCoder = new ethers.utils.AbiCoder();
+      const bridgeDeploymentResult: BridgeDeploymentResult = await loadFixture(deployContracts);
+      proxyBridge = bridgeDeploymentResult.proxyBridge;
+      defaultAdminSigner = bridgeDeploymentResult.defaultAdminSigner;
+    });
+
+    it('Send no data and send some ether to the smart contract', async () => {
```
weiyuan95, line 88:
```suggestion
    it('should increase bridge balance when sending no data and ether to the smart contract', async () => {
```

packages/smartcontracts/src/tests/BridgeProxy.proxyBehaviour.spec.ts
```diff
@@ -0,0 +1,151 @@
...
+    });
+
+    it('Call a method that is not in the implementation interface', async () => {
+      const TestTokenFactory = await ethers.getContractFactory('TestToken');
+      const attachedToken = TestTokenFactory.attach(proxyBridge.address);
+      await expect(attachedToken.totalSupply()).to.reverted;
+    });
+
+    it('Call a method that is in the implementation interface that is not payable, without sending ether', async () => {
+      await defaultAdminSigner.sendTransaction({
```
weiyuan95, line 114: Any particular reason why we are sending transactions this way instead of doing something like `BridgeV1.attach(proxyBridge.address).connect(defaultAdminSigner).addSupportedTokens(...)` 

That way we have strongly typed method calls and it's clearer as well instead of doing native transactions, and it's also still being proxied through the `proxyBridge`, just that we rely on the `BridgeV1` interface in code.

packages/smartcontracts/src/tests/BridgeProxy.proxyBehaviour.spec.ts
```diff
@@ -0,0 +1,151 @@
...
+    it('Call a method that is in the implementation interface that is not payable, and send ether', async () => {
+      await expect(
+        defaultAdminSigner.sendTransaction({
+          to: proxyBridge.address,
+          data:
+            ethers.utils.keccak256(ethers.utils.toUtf8Bytes('addSupportedTokens(address,uint256)')).substring(0, 10) +
+            abiCoder.encode(['address', 'uint256'], [ethers.constants.AddressZero, toWei('10')]).substring(2),
+          value: 100,
+        }),
+      ).to.reverted;
```
weiyuan95, line 132:
```suggestion
      ).to.be.reverted;
```
Is it this instead? 🤔

packages/smartcontracts/src/tests/BridgeProxy.proxyBehaviour.spec.ts
```diff
@@ -0,0 +1,151 @@
...
+  });
+
+  describe('Test different situations of calling the proxy contract', () => {
+    let abiCoder: any;
+    let proxyBridge: BridgeV1;
+    let defaultAdminSigner: SignerWithAddress;
+
+    beforeEach(async () => {
+      abiCoder = new ethers.utils.AbiCoder();
+      const bridgeDeploymentResult: BridgeDeploymentResult = await loadFixture(deployContracts);
```
weiyuan95, line 83:
```suggestion
      const bridgeDeploymentResult = await loadFixture(deployContracts);
```
The return type can be inferred from the function

packages/smartcontracts/src/tests/BridgeProxy.proxyBehaviour.spec.ts
```diff
@@ -0,0 +1,151 @@
...
+      expect(await upgradedProxyBridge.eoaAddressToNonce(accounts[2])).to.equal(1);
+      expect(await upgradedProxyBridge.supportedTokens(testToken.address)).to.equal(true);
+      expect(await upgradedProxyBridge.relayerAddress()).to.equal(defaultAdminSigner.address);
+      expect((await upgradedProxyBridge.tokenAllowances(testToken.address)).dailyAllowance).to.equal(toWei('15'));
+      expect(await upgradedProxyBridge.transactionFee()).to.equal(30);
+    });
+  });
+
+  describe('Test different situations of calling the proxy contract', () => {
+    let abiCoder: any;
```
weiyuan95, line 77:
```suggestion
    let abiCoder: AbiCoder;
```

You'll need to make the appropriate import as well.

packages/smartcontracts/src/tests/BridgeProxy.proxyBehaviour.spec.ts
```diff
@@ -0,0 +1,151 @@
...
+        proxyBridge.bridgeToDeFiChain(ethers.constants.AddressZero, ethers.constants.AddressZero, 0, {
+          value: toWei('10'),
+        }),
+      ).to.emit(proxyBridge, 'BRIDGE_TO_DEFI_CHAIN');
+    });
+
+    it('Call a method that is in the implementation interface that is payable, without sending any ether', async () => {
+      await proxyBridge.addSupportedTokens(ethers.constants.AddressZero, toWei('15'));
+      await expect(
+        proxyBridge.bridgeToDeFiChain(ethers.constants.AddressZero, ethers.constants.AddressZero, 0),
```
weiyuan95, line 147: Hmm, this should not work right? We should not be allowing `0` values? 🤔

weiyuan95 (changes requested):

packages/smartcontracts/src/contracts/NewImplementation.sol
cuongquangnam: Storage layout of smart contracts using inheritance follow the C3 Linearization. 
https://en.wikipedia.org/wiki/C3_linearization
So currently, by using tool like [surya](https://github.com/ConsenSys/surya), we can trace the C3 Linearization of BridgeV1 as follows: 
BridgeV1
  ↖ AccessControlUpgradeable
  ↖ ERC165Upgradeable
  ↖ IERC165Upgradeable
  ↖ IAccessControlUpgradeable
  ↖ ContextUpgradeable
  ↖ EIP712Upgradeable
  ↖ UUPSUpgradeable
  ↖ ERC1967UpgradeUpgradeable
  ↖ Initializable
  ↖ IERC1822ProxiableUpgradeable

The compiler will allocate slots from the bottom of the above chain to the top of the chain (in our case, because IERC1822ProxiableUpgradeable does not have any state variable, the first slot will belong to Initializable contract)

packages/smartcontracts/src/contracts/NewImplementation.sol
cuongquangnam: Yes, you can refer to these: 
https://docs.openzeppelin.com/upgrades-plugins/1.x/proxies
https://docs.soliditylang.org/en/v0.8.17/internals/layout_in_storage.html

packages/smartcontracts/src/tests/BridgeProxy.proxyBehaviour.spec.ts
cuongquangnam: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/ac30219a6a0eb9f98798da39ddef7fe554e43616/contracts/proxy/ERC1967/ERC1967Upgrade.sol#L28

packages/smartcontracts/src/tests/BridgeProxy.proxyBehaviour.spec.ts
cuongquangnam: oh, I think @Abhishekkochar wants to remove the use of tx.wait(), which is fine given the transactions sent out in tests seem to get executed instantly. Made changes in my new commit

packages/smartcontracts/src/tests/BridgeProxy.proxyBehaviour.spec.ts
cuongquangnam: BridgeV1 inherits from ERC1967UpgradeUpgradeable, which is upgradeable version of the above contract.
(https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/blob/v4.8.0/contracts/proxy/ERC1967/ERC1967UpgradeUpgradeable.sol)

weiyuan95: Let's put that as a comment!

packages/smartcontracts/src/tests/BridgeProxy.proxyBehaviour.spec.ts
cuongquangnam: We allow in the current contract

weiyuan95: Definitely should be fixed. @Abhishekkochar can you double check this, and if it's actually an issue we open an issue for it?

packages/smartcontracts/src/tests/BridgeProxy.proxyBehaviour.spec.ts
weiyuan95: Ok let's do it for this, and ignore the test below since it's not possible

cuongquangnam: I think the point is to test different scenarios, can see the analysis in my latest commit

packages/smartcontracts/src/tests/BridgeProxy.proxyBehaviour.spec.ts
```diff
@@ -0,0 +1,171 @@
...
+      expect(receipt.logs[0].topics[1]).to.equal(abiCoder.encode(['address'], [defaultAdminSigner.address]));
+      expect(receipt.logs[0].topics[2]).to.equal(abiCoder.encode(['uint256'], [toWei('1')]));
+      const afterETHBalance = await ethers.provider.getBalance(proxyBridge.address);
+      expect(afterETHBalance.sub(prevETHBalance)).to.equal(toWei('1'));
+    });
+
+    it('Revert when calling a method that is not in the implementation interface', async () => {
+      const TestTokenFactory = await ethers.getContractFactory('TestToken');
+      const attachedToken = TestTokenFactory.attach(proxyBridge.address);
+      await expect(attachedToken.totalSupply()).to.reverted;
```
cuongquangnam, line 130: Can consider adding a test with a function that is non-pure, non-view and not in BridgeV1 ABI, so as to let ethers send out a transaction to the network
https://docs.ethers.org/v5/api/contract/contract/#Contract--metaclass

packages/smartcontracts/src/tests/BridgeProxy.proxyBehaviour.spec.ts
Abhishekkochar: In this case `0` won't be accountable. Token address is `0x0`. If we don't allow `0`, this would result in sending arb value when sending ether.

weiyuan95: Are you referring to the first 2 params? Understand that we have to allow `0` for the first two, but I'm referring to the value of `_amount` here, we can check and make sure that it's non-zero right?

Abhishekkochar: Tracked [here](https://github.com/WavesHQ/bridge/issues/194). Publishing as part of [PR](https://github.com/WavesHQ/bridge/pull/175)

Abhishekkochar: I meant the `_amount` As per the current code, lil change needed for this.

weiyuan95: Resolving this, it is handled with `require(amount > 0)` in #175

packages/smartcontracts/src/tests/BridgeProxy.proxyBehaviour.spec.ts
weiyuan95: Yep, I mean using the method from the contract instance in the `Call a method that is in the implementation interface that is not payable, without sending ether` test, but using the raw calldata string for the `Revert when calling a method that is in the implementation interface that is not payable, and send ether` test since we need to send a `value` which the method call does not allow.

cuongquangnam: Hmm, I would prefer using sendTransaction in this case, as it showcases that the same method can succeed or fail with or without sending Ether. Other tests will still use the contract instance.

weiyuan95 (dismissed):
LGTM! Thanks for the work on this.

@Abhishekkochar any comments on this PR?

packages/smartcontracts/src/tests/BridgeProxy.proxyBehaviour.spec.ts
```diff
@@ -0,0 +1,176 @@
...
+      const eip712Data = {
+        to: accounts[2],
+        amount: toWei('10'),
+        nonce: 0,
+        deadline: ethers.constants.MaxUint256,
+        tokenAddress: testToken.address,
+      };
+
+      const signature = await defaultAdminSigner._signTypedData(domainData, eip712Types, eip712Data);
+      proxyBridge.claimFund(accounts[2], toWei('10'), 0, ethers.constants.MaxUint256, testToken.address, signature);
```
Abhishekkochar, line 48:
```suggestion
     await proxyBridge.claimFund(accounts[2], toWei('10'), 0, ethers.constants.MaxUint256, testToken.address, signature);
```
Missing `await`

packages/smartcontracts/src/tests/BridgeProxy.proxyBehaviour.spec.ts
```diff
@@ -0,0 +1,176 @@
...
+      expect(await upgradedProxyBridge.eoaAddressToNonce(accounts[2])).to.equal(1);
+      expect(await upgradedProxyBridge.supportedTokens(testToken.address)).to.equal(true);
+      expect(await upgradedProxyBridge.relayerAddress()).to.equal(defaultAdminSigner.address);
+      expect((await upgradedProxyBridge.tokenAllowances(testToken.address)).dailyAllowance).to.equal(toWei('15'));
+      expect(await upgradedProxyBridge.transactionFee()).to.equal(30);
+    });
+  });
+
+  describe('Test different situations of calling the proxy contract', () => {
+    let abiCoder: AbiCoder;
```
Abhishekkochar, line 79:
```suggestion
    const  abiCoder = new ethers.utils.AbiCoder();
```
There is no need for this in `beforeEach`

packages/smartcontracts/src/tests/BridgeProxy.proxyBehaviour.spec.ts
```diff
@@ -0,0 +1,176 @@
...
+    //                                      /     \
+    //                       function is payable   \
+    //                                   /        function is not payable
+    //                  succeeds if all goes well       /           \
+    //                                                 /             \
+    //                                          msg.value == 0      msg.value > 0
+    //                                              |                 |
+    //                            succeeds if all goes well         fail
+
+    it('Sucessful when sending no data and send some ether to the smart contract', async () => {
```
Abhishekkochar, line 116: @weiyuan95, I suggest we wait for [PR](https://github.com/WavesHQ/bridge/pull/289) to be merged. We can then remove unneeded test for ether.

Abhishekkochar (changes requested):

packages/smartcontracts/src/tests/BridgeProxy.proxyBehaviour.spec.ts
weiyuan95: Can probably delete this line? It has no effect on the tests (since it still passes when missing an `await`)

packages/smartcontracts/src/tests/BridgeProxy.proxyBehaviour.spec.ts
weiyuan95: Ah. Was this to test the code below?

```
    /**
     * @dev Testing need for this fn. Using to test claim ether.
     */
    fallback() external payable {}

    /**
     * @dev Testing need for this fn. Using to test claim ether.
     */
    receive() external payable {}
```

If yes then yep, we can probably remove since we aren't worrying about ETH anymore.

Abhishekkochar: Yes, most of the tests here testing ether, which we aren't supporting any more.

packages/smartcontracts/src/tests/BridgeProxy.proxyBehaviour.spec.ts
Abhishekkochar: I believe this for checking if the `claimFund` is working properly. It is passing however not sure if the user actually received the tokens.

cuongquangnam: Oh, its purpose is to test whether the nonce after the upgrade is 1 later. Anw, make changes as requested. Thanks for pointing out!

Issue #183, opened 2023-01-12
Implement `EVMTransactionConfirmerQueue`


weiyuan95:
Flow: #176 

#### What would you like to be added:
A transaction confirmer queue that checks that a transaction has reached the necessary number of confirmations on the EVM.

~~If a transaction has not reached the necessary number of confirmations, it will be re-enqueued to be checked again. This does not need to happen instantly though - the job can be delayed by a minute or 2 to allow the chain to mine some blocks.~~

We can keep track of a few variables, the tip T (query-able from chain), last checked block L (needs to be stored in the database coming from https://github.com/WavesHQ/bridge/issues/196 ), and X, which is T - necessaryNumberOfConfirmations (calculated via math, with the necessary number of confirmations coming from a config variable). We will then process the transactions/events from L to X, which will be transactions/blocks with more than the necessary number of confirmations.

Note: the logic stated above might not necessarily need a `Queue`, the dev working on this can implement as they see fit!

There is a potential edge case where a transaction is not able to reach the necessary number of confirmations. Although this is very unlikely, we can set a maximum number of retries, and handle it if it exceeds the maximum number of retries. For eg. 65 confirmations should only take 13 minutes. Assuming that each retry happens every 1 minute, the transaction should definitely be confirmed within 20 retries. If even after 30 retries it's still not confirmed, we will need to handle that particular job, instead of looping it back in to be infinitely retried.

> We will wait for 65 confirmations, unless we want to wait be speedier at the expense of some security. In that case, we will go with 33 confirmations.

#### Why is this needed:
To check that a transaction has reached the necessary number of confirmations on the EVM. This will help us check that the transaction is final, and we can then execute some follow up code.

PR #189, opened 2023-01-12
feat(ConfirmerQueues): Scaffold confirmer queues

### LINKED ISSUE #187: Implement `DFCTransactionConfirmerQueue`
As per title. This is similar to #183 . It might be possible to abstract away some parts of the confirmer queue, especially if we are going to support other chains in the future.

#### Why is this needed:
We need to confirm the transactions on DeFiChain for finality, before executing the follow up mint / burn transactions.

#### Partially Blocked by:
Number of confirmations for DFC is unknown, but this issue can still be started on though.

weiyuan95:
Scaffolds the confirmer queues we will need. These are empty shells - the respective confirmer implementation for DeFiChain and Ethereum will need to be filled in.

There is an example integration with the controller for the DeFiChain one.

Partially fixes #187 and #183

### DIFF
--- a/apps/server/src/queue/DeFiChainTransactionConfirmerQueue.ts
+++ b/apps/server/src/queue/DeFiChainTransactionConfirmerQueue.ts
@@ -1,0 +1,55 @@
+import { Injectable, OnApplicationBootstrap } from '@nestjs/common';
+import { ConfigService } from '@nestjs/config';
+import { Job } from 'bullmq';
+
+import { MultiQueueManager } from './MultiQueueManager';
+import { Queue } from './Queue';
+
+@Injectable()
+export class DeFiChainTransactionConfirmerQueue
+  extends Queue<DeFiChainTransactionConfirmerJob, any>
+  implements OnApplicationBootstrap
+{
+  constructor(
+    readonly queue: MultiQueueManager<DeFiChainTransactionConfirmerJob, void>,
+    private readonly configService: ConfigService,
+  ) {
+    super(configService.getOrThrow('queues.defichain'), queue);
+  }
+
+  async onApplicationBootstrap(): Promise<void> {
+    await super.startConsumers();
+  }
+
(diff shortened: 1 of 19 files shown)


package.json
```diff
@@ -11,7 +11,7 @@
     "test": "turbo run test",
     "test:hardhat": "turbo run test:hardhat --filter=./packages/smartcontracts",
     "dev": "turbo run dev --parallel",
-    "lint": "turbo run lint",
+    "lint": "turbo run lint -- --fix",
```
weiyuan95, line 14: We can `fix` the errors that the linter throws. Some of them, at least.

weiyuan95:
Setting these as draft for now - we might not need this.

Also, should use the redis containers from `sticky` instead of rolling my own implementation.

weiyuan95:
Setting this to draft for now - it might now be necessary given our current implementation.

apps/server/src/queue/DeFiChainTransactionConfirmerQueue.ts
```diff
@@ -0,0 +1,55 @@
...
+    if ((await this.isTransactionConfirmed(transactionId)) === false) {
+      // Throw error here to fail job and let bull auto retry again after X seconds delay
+      throw new Error();
+    }
+  }
+
+  // TODO: Logic to check the number of confirmations
+  // eslint-disable-next-line  @typescript-eslint/no-unused-vars
+  async isTransactionConfirmed(transactionId: string): Promise<boolean> {
+    return true;
```
thedoublejay, line 49: The confirmer logic can be used on this service - https://github.com/WavesHQ/bridge/pull/250/files#diff-5eca400a67123ad45b8bf1e241fcc7c8160b8cadf4675c847b6288897124ab2bR57:~:text=waitForTxConfirmation

thedoublejay:
@weiyuan95 can we close this

PR #204, opened 2023-01-17
feature(api): setup postgresql for db connectivity

### LINKED ISSUE #196: Setup DB (Postgres)

fullstackninja864:
In this PR we are setting up postgresql for db connectivity. We used Prisma as a ORM

Fixes #196

#### Additional comments?:
Why Prisma and not Type ORM ?
Prisma generates a lightweight database client that exposes a tailored and fully type-safe API to 
read and write data for the models that are defined in the Prisma schema, following the DataMapper 
ORM pattern rather than Active Record. also prisma's DSL for data modeling is lean, simple and intuitive to use

Commands
1. Initial setup
Assuming your local Postgres server is still up and running 
setup `.env` file in [/server](https://github.com/WavesHQ/bridge/tree/main/apps/server) folder  and update db configuration
```
  POSTGRES_DB=<db name>
  POSTGRES_USER=<db user>
  POSTGRES_PASSWORD=<db password>
  DATABASE_URL="postgresql://${POSTGRES_USER}:${POSTGRES_PASSWORD}@localhost:5432/${POSTGRES_DB}?schema=public"
```
Run > ```pnpm run migration:dev```
To create initial setup locally you need to run above command. this will generate DB table 

2. Generate artifacts 
Run > ```pnpm run prisma:generate```
The generate command generates assets like Prisma Client based on the generator and data model blocks defined in your prisma/schema.prisma file
The way Prisma Client works is that it auto-generates the database client code for you into the node_modules folder of your project.
We need to add this command in deployment pipeline as auto-generation won't happen because we will run npm ci before the step to copy over the files in our local directory. Therefore, we need a command to do the auto-generation manually.
we have to run this command every time when me do changes in db schema

3. Create new migration 
Run > ```pnpm run migration:create```
When we make any change in db schema we need to create migration. In this case above command will create migration file without applying it so that we can update DB migration command manually to avoid any data loss and after updating the migration we can run ```pnpm run migration:dev```

4. Deploy migration on prod
Run > ```pnpm run migration:deploy```
The migrate deploy command applies all pending migrations, and creates the database if it does not exist. Primarily used in non-development environments.

5. Prisma Client
To interact with DB we need to install `@prisma/client`.
example 
``` 
  import { PrismaClient } from '@prisma/client';
  // create client 
  const prisma = new PrismaClient();

  async function example() {
    // add entry in table 
    await prisma.pathIndex.create({
        data: {
          index: 1,
          address: 'dsgfhjklsdhgajsgdgjaagdsjags',
      },
    });
    // list all data from table
    const data = await prisma.pathIndex.findMany();
    console.dir(data, { depth: null });
  }
```
#### Developer Checklist:

- [ ] Unit tests\*
- [ ] Added e2e tests\*
- [ ] Added translations\*


thedoublejay:
@fullstackninja864 remember to update the docs to pnpm instead of npm

fullstackninja864:
@weiyuan95 Currently we don't have any scenario for which we have to write down test case for.  we can implement  test container for integration tests when required in future

weiyuan95:
> Quick comment from me about this - although this is great from a front-end/ui POV (running a script that runs the `docker-compose`, and any necessary UI scripts), it does not address the fact that we cannot use this to run integration tests from the server (testing that the server related components are integrated correctly).
> 
> To get around that, we can probably provision a container for `postgres`, similar to the [`HardhatContainer`](https://github.com/WavesHQ/bridge/blob/main/packages/smartcontracts/src/containers/HardhatNetworkContainer.ts) and use that for our integration tests.

Ignore me on this, the postgres container is already provisioned from `sticky-testcontainers`

thedoublejay (approved):

PR #211, opened 2023-01-18
feature(api): setup api rate limiter to avoid spamming

### LINKED ISSUE #199: Rate limiter for endpoints
To prevent spamming and to ensure our endpoints are protected, we need to implement rate limiting on our endpoints.

fullstackninja864:
In this PR to prevent spamming and to ensure our endpoints are protected, we have added rate limiting on our endpoints.
This will ensure API requests made by a throttled user or app will fail after reaching its limit 

Fixes #199

#### Additional comments?:
Default rate limiting is 10 requests from the same IP can be made to a single endpoint in 1 minute. This will get reset after 1 min interval.

Decoraters
1. API specific custom rate limiting 
we can use ```@Throttle(limit: number = 30, ttl: number = 60)```  to define api specific rate limiting  

2. Skip rate limiting
To skip the rate limiting for specific api we can use  ```@SkipThrottle(skip = true)```

This decorator can be used to skip a route or a class or to negate the skipping of a route in a class that is skipped.
```
@SkipThrottle()
@Controller()
export class AppController {
  @SkipThrottle(false)
  dontSkip() {}

  doSkip() {}
}
```

#### Developer Checklist:

- [ ] Unit tests\*
- [ ] Added e2e tests\*
- [ ] Added translations\*

### DIFF
--- a/apps/server/package.json
+++ b/apps/server/package.json
@@ -39,6 +39,7 @@
     "@nestjs/config": "^2.2.0",
     "@nestjs/core": "^9.2.1",
     "@nestjs/platform-fastify": "^9.2.1",
+    "@nestjs/throttler": "^3.1.0",
     "@waveshq/walletkit-core": "^0.20.0",
     "class-transformer": "^0.5.1",
     "class-validator": "^0.14.0",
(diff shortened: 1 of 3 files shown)


lykalabrada (approved):
LGTM 👍

thedoublejay (approved):

Issue #159, opened 2022-12-30
Contracts should be upgradable


Abhishekkochar:
#### What would you like to be added:
The bridge contract should be upgradable, in the event that we need to update any of the logic, or if a vulnerability was found in the contract.

This should keep in mind any of the usual caveats, for eg. a contract should only be upgradable by an admin address, etc.

Abhishekkochar:
@weiyuan95, In regards to recent upgrade (V2) we can possibly close this issue.

weiyuan95:
Contracts have already been upgraded, and we've addressed the teething issues that came with it as well.

PR #212, opened 2023-01-18
feat(SmartContracts): Deployment script for Goerli test net

Abhishekkochar:
This PR introduces the deployment script for the Bridge and test tokens smart contracts

Fixes #

#### Developer Checklist:

- [ ] Unit tests\*
- [ ] Added e2e tests\*
- [ ] Added translations\*

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,3 +10,21 @@
 DeFiChain ERC-20 Bridge
 
 TODO
+
+## Deployed Smart Contracts on Goerli testnet
+
+### MUSDT
+
+MUSDT Contract address: [0xA218A0EA9a888e3f6E2dfFdf4066885f596F07bF](https://goerli.etherscan.io/address/0xA218A0EA9a888e3f6E2dfFdf4066885f596F07bF)
+
+### MUSDC
+
+MUSDC Contract address: [0xB200af2b733B831Fbb3d98b13076BC33F605aD58](https://goerli.etherscan.io/address/0xB200af2b733B831Fbb3d98b13076BC33F605aD58)
+
+### BridgeV1
+
+BridgeV1 Contract address: [0xE029B5156c2e597c72f7c8D279411e1fD9a30126](https://goerli.etherscan.io/address/0xE029B5156c2e597c72f7c8D279411e1fD9a30126)
+
+### BridgeProxy
+
+BridgeProxy Contract addrress: [0x93fE70235854e7c97A5db5ddfC6eAAb078e99d3C](https://goerli.etherscan.io/address/0x93fE70235854e7c97A5db5ddfC6eAAb078e99d3C)
--- a/README.md
+++ b/README.md
@@ -19,7 +19,11 @@ To deploy ERC20 token user will have to run a command `npx hardhat run --network
 
 To verify the said tokens and other contracts, there would be a prompt on terminal after running the deployment command that devs will need to run after.
 
-Devs need to deploy the `BridgeV1` contract contract before the `BridgeProxy`.
+Devs need to deploy the `BridgeV1` implementation contract before the `BridgeProxy`.
+
+`BridgeProxy` should only need to be deployed _once_ to a blockchain. Subsequent deployments should only be deploying the implementation contract (`BridgeV2`, `BridgeV3`, etc), before calling `_upgradeTo` of the `BridgeProxy` contract.
+
+This follows the [proxy pattern](https://blog.openzeppelin.com/proxy-patterns/), with the behaviour being inherited from `OpenZeppelin` proxy contracts.
 
 `BridgeV1` can be deployed with the command `npx hardhat run --network goerli ./scripts/deployBridgeImplementation.ts`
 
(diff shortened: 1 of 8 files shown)


packages/smartcontracts/src/.env.example
```diff
@@ -0,0 +1,3 @@
+PRIVATE_KEY = '123test123test123test123test123test123test123test'
```
weiyuan95, line 1: How do we provide a private key for this, if we use a wallet like metamask?

packages/smartcontracts/src/hardhat.config.ts
```diff
@@ -82,6 +87,11 @@ const config: HardhatUserConfig = {
       // could be larger than the stipulated max size in EIP-170
       allowUnlimitedContractSize: true,
     },
+    goerli: {
+      url: process.env.GOERLI_URL || '',
+      accounts: process.env.PRIVATE_KEY !== undefined ? [process.env.PRIVATE_KEY] : [],
+      gasPrice: 30000000000, // this is 30 Gwei
```
weiyuan95, line 93: What is this `gasPrice` for?

weiyuan95 (changes requested):
Some quick comments, the rest LGTM for now, will do a more detailed review later.

CI is failing as well, i'm guessing you have dotenv installed globally. Need to add it to the package.json

packages/smartcontracts/src/.env.example
Abhishekkochar: Private key will be stored in `.env`. We will be using a script instead of using MetaMask. We can use MetaMask if using  remix.

packages/smartcontracts/src/hardhat.config.ts
Abhishekkochar: This is for the tx. If the `gasPrice` is too low, tx will get stuck in Mempool.

Abhishekkochar:
> Some quick comments, the rest LGTM for now, will do a more detailed review later.
> 
> CI is failing as well, i'm guessing you have dotenv installed globally. Need to add it to the package.json

Already on it.

packages/smartcontracts/src/.env.example
weiyuan95: Ah. I was referring to [this](https://metamask.zendesk.com/hc/en-us/articles/360015289632-How-to-export-an-account-s-private-key) but I found it after a quick google 🤣

Abhishekkochar: Oh ya hahah. That's how I do it as well.

Abhishekkochar: I have added the above link in .env.example

README.md
```diff
@@ -13,6 +13,21 @@ TODO
 
 ## Deployed Smart Contracts on Goerli testnet
 
+## Deploy ERC20 tokens 'MUSDT' & 'MUSDC'
+
+To deploy ERC20 token user will have to run a command `npx hardhat run --network goerli ./scripts/deployERC20.ts` in smartContract directory.
+
+To verify the said tokens and other contracts, there would be a prompt on terminal after running the deployment command that devs will need to run after.
+
+Devs need to deploy the `BridgeV1` contract contract before the `BridgeProxy`.
```
weiyuan95, line 22:
```suggestion
Devs need to deploy the `BridgeV1` implementation contract before the `BridgeProxy`.

`BridgeProxy` should only need to be deployed _once_ to a blockchain. Subsequent deployments should only be deploying the implementation contract (`BridgeV2`, `BridgeV3`, etc), before calling `_upgradeTo` of the `BridgeProxy` contract.

This follows the [proxy pattern](https://blog.openzeppelin.com/proxy-patterns/), with the behaviour being inherited from `OpenZeppelin` proxy contracts.
```

Looking at this, I realised that we have not exposed `_upgradeTo` on `BridgeProxy` yet, but I'm guessing that will be part of #159

weiyuan95 (changes requested):
One last quick point 😬

README.md
Abhishekkochar: https://goerli.etherscan.io/address/0x93fE70235854e7c97A5db5ddfC6eAAb078e99d3C#writeProxyContract. I have Updated the proxy pattern (will now show as proxy as contract.) We can upgrade this anytime as long as it's satisfy `onlyProxy`.

weiyuan95: Hmm, i'm not sure what you mean. We have 2 proxy contracts now?

https://goerli.etherscan.io/address/0x93fE70235854e7c97A5db5ddfC6eAAb078e99d3C#readProxyContract 
https://goerli.etherscan.io/address/0x50894DE4589BaC6FBA9D77e3c988Ff5594E6710A#readProxyContract

They both look the same to me 🤔 . Also, I thought that we had to expose `_upgradeTo` since it's currently an internal method.

We can handle those issues in other PRs though, will merge this in first so that we can work on the utility and fixture scripts for UI devs.

weiyuan95: Update to this
- `upgradeTo` is already handled in one of the open zep contract the bridge implements
- The other proxy contract was probably from Cuong testing the deployment

PR #180, opened 2023-01-11
feature(core): generate unique address in server/defichain

pierregee:
#### What this PR does / why we need it:
- [x] Generate unique address in API
- [ ] Integrate endpoint in Bridge UI 
#### Which issue(s) does this PR fixes?:

Fixes #

#### Additional comments?:
- Since the private key for the BR consortium is [not provided yet](https://github.com/WavesHQ/bridge/pull/180/files#diff-2e5a29eaf834827c8bf61cc0435553c551000693902eb0f33298ed08fed3d55aR18), generated a new wallet first to at least start the integration with Bridge UI and prepare the bulk of code on generating unique address 

#### Developer Checklist:

- [ ] Read your code changes at least once
- [ ] No console errors on web
- [ ] Tested on Light mode and Dark mode\*
- [ ] Your UI implementation visually matched the rendered design\*
- [ ] Unit tests\*
- [ ] Added e2e tests\*
- [ ] Added translations\*

PR #223, opened 2023-01-19
feature(apps/server): initialize whale-wallet-* for ops on DeFiChain

### LINKED ISSUE #144: `Server` - `DeFiChain` - Address Generator Endpoint
Create an endpoint to generate an address per transaction when doing DeFiChain -> ERC20. Should be able to optionally specify network. If not provided, set to mainnet as default.

### LINKED ISSUE #193: DeFiChain Wallet Initializer
Initialize wallet that will be used to sign transactions. Private keys handling will be on a separate ticket.

thedoublejay:
This PR creates the ability to initialize a DeFiChain Whale Wallet on Bridge. I specifically used "Whale Wallet" in case we use different implementations of wallet in the future.

To explain the workflow:

1. It fetches the mnemonic of BR from the config. This will be injected during deployment. 
2. `WhaleWalletProvider` will fetch the data on config and create the wallet on every request. Wallets are tied to network (Mainnet, testnet, regtest). 
3. All consumers of WhaleWalletProvider (in this PR, the address generator), will just reference this and will be able to do ops on DeFiChain.

For testing (playground), we specified a static set of mnemonic. This Mnemonic has been defined in Playground to be a member of Consortium so we can simulate the full e2e experience.

This PR on Jellyfish repo https://github.com/JellyfishSDK/jellyfish/pull/1981, will allow us to set the mnemonic used on bridge to be a part of Consortium. So everytime the bridge mints via the consortium, we will able to verify it on playground network.

Linked to https://github.com/WavesHQ/bridge/pull/180/files

Fixes #193 
Fixes #144 

#### Developer Checklist:

- [ ] Unit tests\*
- [ ] Added e2e tests\*
- [ ] Added translations\*


lykalabrada (approved):

pierregee (approved):

apps/server/docker-compose.yml
```diff
@@ -17,7 +17,7 @@ services:
       - '/var/run/docker.sock:/var/run/docker.sock:ro'
 
   defi-blockchain:
-    image: defi/defichain:3.2.1
+    image: defi/defichain:3.2.2
```
weiyuan95, line 20: https://docs.renovatebot.com/modules/manager/docker-compose/

Renovate can handle these as well, to make it automated 😬

apps/server/src/defichain/providers/WhaleWalletProvider.ts
```diff
@@ -0,0 +1,73 @@
...
+import { ConfigService } from '@nestjs/config';
+import { EnvironmentNetwork, getBip32Option, getJellyfishNetwork } from '@waveshq/walletkit-core';
+import { WalletPersistenceDataI, WalletType } from '@waveshq/walletkit-ui';
+
+import { SupportedNetwork } from '../model/NetworkDto';
+import { WhaleApiService } from '../services/WhaleApiService';
+
+@Injectable()
+export class WhaleWalletProvider {
+  constructor(private readonly whaleClient: WhaleApiService, private readonly configService: ConfigService) {}
```
weiyuan95, line 15: To properly leverage the `class` scope, we can do something like `this.mappedNetwork = mapNetwork(network)` and `this.mnemonic = ...` so that we don't need to pass in the same parameter to the class methods.

A good indicator on when something should be a class property is when the same property is passed to all the class methods as a parameter. In this case `network` is definitely a clear contender.

apps/server/src/defichain/providers/WhaleWalletProvider.ts
```diff
@@ -0,0 +1,73 @@
...
+
+  createWallet(network: SupportedNetwork = SupportedNetwork.mainnet) {
```
weiyuan95, line 17: Missing the return type here.

apps/server/src/defichain/providers/WhaleWalletProvider.ts
```diff
@@ -0,0 +1,73 @@
...
+      raw: data,
+    };
+  }
+
+  private initJellyfishWallet<HdNode extends WalletHdNode>(
+    provider: WalletHdNodeProvider<HdNode>,
+    network: EnvironmentNetwork,
+    client: WhaleApiClient,
+  ): JellyfishWallet<WhaleWalletAccount, HdNode> {
+    const accountProvider = new WhaleWalletAccountProvider(client, getJellyfishNetwork(network));
```
weiyuan95, line 55: `client` here can be `this.whaleClient.getClient(network)` directly.

apps/server/src/defichain/services/WhaleWalletService.ts
```diff
@@ -0,0 +1,26 @@
...
+import { SupportedNetwork } from '../model/NetworkDto';
+import { WhaleWalletProvider } from '../providers/WhaleWalletProvider';
+
+@Injectable()
+export class WhaleWalletService {
+  constructor(private readonly whaleWalletProvider: WhaleWalletProvider) {}
+
+  async generateAddress(network: SupportedNetwork = SupportedNetwork.mainnet): Promise<string> {
+    const wallet = this.whaleWalletProvider.createWallet(network);
+    return wallet.getAddress().catch((error: Error) => {
```
weiyuan95, line 12: We can use a `try { const address = wallet.getAddress() } catch { }` here instead of using `.catch` to leverage on `await`

apps/server/docker-compose.yml
thedoublejay: It's handled but needed it for this PR

apps/server/src/defichain/providers/WhaleWalletProvider.ts
thedoublejay: Hmmm this only works if we don't change the network. However, right now it's designed for users to be able to change network on every request.

PR #242, opened 2023-01-26
feat(Contracts): Local and Mainnet deployment scripts

### LINKED ISSUE #236: Fix flaky `SetupLocalTestTask` spec
#### What happened:
To improve the devx on the smart contract integration side, the `SetupLocalTest` task was quickly implemented and pushed for speed. However, the `SetupLocalTestTask` spec is flaky, for the reasons outlined [here](https://github.com/WavesHQ/bridge/pull/224#discussion_r1082075628)

It's inherently flaky on the CI due to the asynchronous nature in which we run tests, as the contract address can change.

The test should be passing deterministically, with the contract addresses being returned from the task.

#### How to reproduce it (as minimally and precisely as possible):
It fails on the CI from time to time.

Abhishekkochar:
This fixes the current hardhat task issue. Two instance of the contracts were getting deployed when running the hardhat task. User can run the following command to mint mock tokens and approve the proxy contract `npx hardhat run --network development ./scripts/localContractsDeployment.ts`

This will only work on localhost/hardhat

Fixes #236 

#### Developer Checklist:

- [ ] Unit tests\*
- [ ] Added e2e tests\*
- [ ] Added translations\*


packages/smartcontracts/src/hardhat.config.ts
```diff
@@ -162,6 +91,13 @@ const config: HardhatUserConfig = {
       url: process.env.GOERLI_URL || '',
       accounts: process.env.PRIVATE_KEY !== undefined ? [process.env.PRIVATE_KEY] : [],
       gasPrice: 30000000000, // this is 30 Gwei
+      chainId: 5,
+    },
+    mainnet: {
+      url: process.env.MAINNET_URL || '',
+      accounts: process.env.PRIVATE_KEY !== undefined ? [process.env.PRIVATE_KEY] : [],
+      gasPrice: 30000000000, // this is 30 Gwei
```
weiyuan95, line 99: Is this necessary instead of letting this value be automatically decided? Or is this just to expedite the deployment? 🤔

packages/smartcontracts/src/scripts/deployBridgeProxy.ts
```diff
@@ -1,34 +1,46 @@
...
 
-main().catch((error) => {
-  console.error(error);
-  process.exitCode = 1;
-});
+interface InputAddresses {
+  AdminAddress: string;
+  OperationalAddress: string;
+  RelayerAddress: string;
+  BridgeV1Address: string;
```
weiyuan95, line 41:
```suggestion
  adminAddress: string;
  operationalAddress: string;
  relayerAddress: string;
  bridgeV1Address: string;
```

packages/smartcontracts/src/hardhat.config.ts
Abhishekkochar: We can let this be as it is. In some scenarios, we will need to set higher gas price if the network is congested.

packages/smartcontracts/src/scripts/deployERC20.ts
```diff
@@ -1,23 +1,31 @@
...
+    );
+  }
+  return { usdtContract: mockTokenUSDT, usdcContract: mockTokenUSDC };
 }
 
-main().catch((error) => {
-  console.error(error);
-  process.exitCode = 1;
-});
+interface TestTokenAddress {
```
weiyuan95, line 28:
```suggestion
interface TestTokens {
```

packages/smartcontracts/src/tests/LocalTest.spec.ts
```diff
@@ -0,0 +1,35 @@
...
+    // Check if the accounts[0] has the admin role.
+    const DEFAULT_ADMIN_ROLE = '0x0000000000000000000000000000000000000000000000000000000000000000';
+    expect(await proxyBridge.hasRole(DEFAULT_ADMIN_ROLE, adminAddress)).to.equal(true);
+    // Check if the relayer address is same as accounts[0]
+    expect(adminAddress).to.be.equal(await proxyBridge.relayerAddress());
+    // Check if the accounts[1] has the OPERATIONAL_ROLE.
+    const OPERATIONAL_ROLE = ethers.utils.solidityKeccak256(['string'], ['OPERATIONAL_ROLE']);
+    expect(await proxyBridge.hasRole(OPERATIONAL_ROLE, operationalSigner.address)).to.equal(true);
+  });
+});
```
weiyuan95, line 35:
```suggestion
describe('SetupLocalTestTask', () => {
  it('should set up the expected local testnet state', async () => {
    // Given that the setupLocalTest task is run
    const { usdtContract, usdcContract, bridgeProxy } = await mintAndApproveTestTokensLocal();

    const expectedTimeStamp = Math.floor(Date.now() / 1000);

    // suppressing type error - method is actually properly typed
    // @ts-ignore
    const [adminSigner, operationalSigner] = await ethers.getSigners();
    const adminAddress = adminSigner.address;
    const operationalAddress = operationalSigner.address;

    const musdtContract = TestToken__factory.connect(usdtContract.address, adminSigner);
    const musdtAddress = musdtContract.address;
    const musdcContract = TestToken__factory.connect(usdcContract.address, adminSigner);
    const musdcAddress = musdcContract.address;
    // behind proxy, so we need to use the proxy address
    const proxyBridge = BridgeV1__factory.connect(bridgeProxy.bridgeProxy.address, adminSigner);

    // When checking the ERC20 balances of the EOA
    expect(await musdtContract.balanceOf(adminAddress)).to.equal(toWei('100000'));
    expect(await musdcContract.balanceOf(adminAddress)).to.equal(toWei('100000'));
    // When checking the admin address of the bridge
    // Then the EOA account should have the admin role.
    const DEFAULT_ADMIN_ROLE = '0x0000000000000000000000000000000000000000000000000000000000000000';
    expect(await proxyBridge.hasRole(DEFAULT_ADMIN_ROLE, adminAddress)).to.equal(true);

    // When checking the operational address of the bridge
    // Then the EOA account should have the operational role.
    const OPERATIONAL_ROLE = ethers.utils.solidityKeccak256(['string'], ['OPERATIONAL_ROLE']);
    expect(await proxyBridge.hasRole(OPERATIONAL_ROLE, operationalAddress)).to.equal(true);

    // When checking that MUSDT is supported on the bridge
    // Then the token should be supported
    expect(await proxyBridge.supportedTokens(musdtAddress)).to.equal(true);
    const musdtSupportedTokenInfo = await proxyBridge.tokenAllowances(musdtAddress);

    // And the token should have the expected reset time
    expect(musdtSupportedTokenInfo[0].eq(expectedTimeStamp)).to.equal(true);

    // And the token should have the expected daily allowance
    expect(musdtSupportedTokenInfo[1].eq(ethers.constants.MaxInt256)).to.equal(true);

    // When checking that MUSDC is supported on the bridge
    // Then the token should be supported
    expect(await proxyBridge.supportedTokens(musdcAddress)).to.equal(true);
    const musdcSupportedTokenInfo = await proxyBridge.tokenAllowances(musdcAddress);

    // And the token should have the expected reset time
    expect(musdcSupportedTokenInfo[0].eq(expectedTimeStamp)).to.equal(true);

    // And the token should have the expected daily allowance
    expect(musdcSupportedTokenInfo[1].eq(ethers.constants.MaxInt256)).to.equal(true);

    // When checking that bridgeToDeFiChain can be called with a MUSDT token
    // And provisioning the necessary approval
    await musdtContract.approve(proxyBridge.address, ethers.constants.MaxInt256);

    // Then the call should not revert
    await proxyBridge.bridgeToDeFiChain(ethers.constants.AddressZero, musdtAddress, 1);

    // And the EOA's balance of MUSDT should be reduced by 1
    expect((await musdtContract.balanceOf(adminAddress)).eq(999_999_999)).to.eq(true);

    // When checking that bridgeToDeFiChain can be called with a MUSDC token
    // And provisioning the necessary approval
    await musdcContract.approve(proxyBridge.address, ethers.constants.MaxInt256);

    // Then the call should not revert
    await proxyBridge.bridgeToDeFiChain(ethers.constants.AddressZero, musdtAddress, 1);

    // And the EOA's balance of MUSDT should be reduced by 1
    expect((await musdcContract.balanceOf(adminAddress)).eq(999_999_999)).to.eq(true);
  });
});
```

IMO this is the ideal test case that we should aim towards since we want to add the tokens support for the devs as well. I think we'll need to make the changes in `localContractsDeployment.ts` to account for it.

weiyuan95 (changes requested):
Some quick comments. The package's README.md will also need to be changed to account for using the script instead of the hardhat task.

packages/smartcontracts/src/hardhat.config.ts
weiyuan95: This is [paying almost double](https://etherscan.io/gastracker) of the current gas price, but since it's a one time thing i'm not too concerned 🤣 . Would rather get the contract deployed.

Abhishekkochar: Screenshot 2023-01-27 at 3 01 11 PM
This is the deployment cost for current SMs.

README.md
```diff
@@ -127,3 +127,16 @@ BridgeV1 Contract address: [0xE029B5156c2e597c72f7c8D279411e1fD9a30126](https://
...
+To deploy the Smart Contracts on local testnet, devs can run the command `npx hardhat run --network hardhat ./scripts/localContractsDeployment.ts` in smartContract directory.
+
+This script will deploy all needed contracts. Will mint `100,000` MUSDC and MUSDT token to the user(in this case, accounts[0]).This will also approve the Bridge contract and add test tokens as supported tokens for bridging with maximum daily allowance.
+
+Following addresses will be the admin, operational roles and relayer address.
+Admin == accounts[0],
+Operational == accounts[1],
+Relayer address == accounts[0]
+
+Devs can change these addresses as per their requirements in `../scripts/localContractsDeployment.ts` file.
```
weiyuan95, line 145: Would shift this section to `packages/smartcontracts/README.md` since it does not really concern the Bridge repo as a whole and can be scoped to the package.

packages/smartcontracts/src/scripts/localContractsDeployment.ts
```diff
@@ -0,0 +1,55 @@
...
+  const defaultAdminSigner = await ethers.getSigner(accounts[0]);
+  const defaultOperationalSigner = await ethers.getSigner(accounts[1]);
+  const { bridgeV1 } = await bridgeImplementation();
+  const { bridgeProxy } = await deployBridgeProxy({
+    adminAddress: defaultAdminSigner.address,
+    operationalAddress: defaultOperationalSigner.address,
+    relayerAddress: defaultAdminSigner.address,
+    bridgeV1Address: bridgeV1.address,
+  });
+  const bridgeImplementationContract = bridgeV1.attach(bridgeProxy.address);
```
weiyuan95, line 22: We should be logging
- The EOA address that receives the funds (ie. signers[0].address) so that it's easy for the devs to know what address to use
- The address of the proxy that the UI should interact with

packages/smartcontracts/src/hardhat.config.ts
weiyuan95: Are those the units of gas used?

But ok, we can review the gas price again before deploying to mainnet

weiyuan95 (changes requested):

packages/smartcontracts/README.md
```diff
@@ -2,15 +2,18 @@
...
 A package which contains the Ethereum smart contracts for the DeFiChain to Ethereum bridge.
 
-## Setting up local environment
+## Deployed Smart Contracts on Hardhat/Local testnet
 
-This section outlines how to get started on integrating with the smart contracts for the Bridge on your dApp.
-There are many different ways, but this is the simplest and fastest method.
+To deploy the Smart Contracts on local testnet, devs can run the command `npx hardhat run --network hardhat ./scripts/localContractsDeployment.ts`.
 
-1. In your terminal, run `npx hardhat --config ./src/hardhat.config.ts node` from this package's directory
```
weiyuan95, line 10: We still need to leave this in so that people will know how to start a node 😅

Abhishekkochar: We don't need to start a node for this.  This can be done without running the node first

weiyuan95: Running just the script as is, there won't be a node available at `localhost:8545` right? 🤔 That's what I remember from running the task previously.

ie. If i just run `npx hardhat run --network hardhat ./scripts/localContractsDeployment.ts`, I can't actually access the local net?

packages/smartcontracts/src/scripts/localContractsDeployment.ts
weiyuan95: Think this is still missing, you didn't add the necessary `console.log`s

packages/smartcontracts/README.md
Abhishekkochar: Yes, you are right. I was thinking from the test perspective. Fixed

Abhishekkochar: When switching to localHost in metamask, TestToken addresses are invalid. Investigating this for now.

packages/smartcontracts/src/scripts/localContractsDeployment.ts
lykalabrada: Hello @Abhishekkochar It's working fine on my end, but as mentioned by @weiyuan95 please also log bridge proxy address. . Thank you!

lykalabrada (dismissed):
Tested from my end, works fine 👍

weiyuan95 (approved):
Approving after applying some code suggestions

lykalabrada (approved):

PR #256, opened 2023-01-30
feat(EVMTransactionConfirmer): Build getAllEventsFromBlockNumber function and integration tests

### LINKED ISSUE #183: Implement `EVMTransactionConfirmerQueue`

greggtan:
The getAllEventsFromBlockNumber function returns an array of confirmed BRIDGE_TO_DEFI_CHAIN events from a given block number. It will be used for the backend server to bridge EVM tokens to DFC.

Fixes #183

#### Additional comments: 
It is a POST request that takes in blockNumber & contractAddress as body params. 

#### Developer Checklist:

- [x] Unit tests\*
- [x] Added e2e tests\*
- [ ] Added translations\*

### DIFF
--- a/apps/server/src/AppConfig.ts
+++ b/apps/server/src/AppConfig.ts
@@ -12,10 +12,28 @@ export function appConfig() {
       rpcUrl: process.env.ETHEREUM_RPC_URL || 'localhost:8545',
     },
     contract: {
-      address: process.env.BRIDGE_CONTRACT_ADDRESS || '',
+      bridgeProxy: {
+        mainnetAddress: undefined,
+        testnetAddress: '0x93fE70235854e7c97A5db5ddfC6eAAb078e99d3C',
+      },
     },
   };
 }
+export type AppConfig = {
+  defichain?: {
+    mainnet?: string;
+    regtest?: string;
+  };
+  ethereum?: {
+    rpcUrl?: string;
+  };
+  contract?: {
+    bridgeProxy?: {
+      mainnetAddress?: string;
+      testnetAddress?: string;
+    };
+  };
+};
 
 export const ENV_VALIDATION_SCHEMA = Joi.object({
   ETHEREUM_RPC_URL: Joi.string().ip(),
--- a/apps/server/src/AppConfig.ts
+++ b/apps/server/src/AppConfig.ts
@@ -11,6 +11,9 @@ export function appConfig() {
     ethereum: {
       rpcUrl: process.env.ETHEREUM_RPC_URL || 'localhost:8545',
     },
+    contract: {
+      address: process.env.BRIDGE_CONTRACT_ADDRESS || '',
+    },
   };
 }
 
--- a/apps/server/src/AppConfig.ts
+++ b/apps/server/src/AppConfig.ts
@@ -20,9 +20,11 @@ export function appConfig() {
   };
 }
 
-type DeepPartial<T> = {
-  [K in keyof T]?: T[K] extends object ? DeepPartial<T[K]> : T[K];
-};
+type DeepPartial<T> = T extends object
+  ? {
+      [P in keyof T]?: DeepPartial<T[P]>;
+    }
+  : T;
 export type AppConfig = DeepPartial<ReturnType<typeof appConfig>>;
 
 export const ENV_VALIDATION_SCHEMA = Joi.object({
--- a/apps/server/src/AppConfig.ts
+++ b/apps/server/src/AppConfig.ts
@@ -19,21 +19,11 @@ export function appConfig() {
     },
   };
 }
-export type AppConfig = {
-  defichain?: {
-    mainnet?: string;
-    regtest?: string;
-  };
-  ethereum?: {
-    rpcUrl?: string;
-  };
-  contract?: {
-    bridgeProxy?: {
-      mainnetAddress?: string;
-      testnetAddress?: string;
-    };
-  };
+
+type DeepPartial<T> = {
+  [K in keyof T]?: T[K] extends object ? DeepPartial<T[K]> : T[K];
 };
+export type AppConfig = DeepPartial<ReturnType<typeof appConfig>>;
 
 export const ENV_VALIDATION_SCHEMA = Joi.object({
   ETHEREUM_RPC_URL: Joi.string().ip(),
--- a/apps/server/src/AppService.ts
+++ b/apps/server/src/AppService.ts
@@ -1,4 +1,5 @@
 import { Inject, Injectable } from '@nestjs/common';
+import { ConfigService } from '@nestjs/config';
 import { BigNumber, ethers } from 'ethers';
 import { BridgeV1__factory } from 'smartcontracts';
 
@@ -6,7 +7,10 @@ import { ETHERS_RPC_PROVIDER } from './modules/EthersModule';
 
 @Injectable()
 export class AppService {
-  constructor(@Inject(ETHERS_RPC_PROVIDER) readonly ethersRpcProvider: ethers.providers.StaticJsonRpcProvider) {}
+  constructor(
+    @Inject(ETHERS_RPC_PROVIDER) readonly ethersRpcProvider: ethers.providers.StaticJsonRpcProvider,
+    private configService: ConfigService,
+  ) {}
 
   async getBlockHeight(): Promise<number> {
     return this.ethersRpcProvider.getBlockNumber();
@@ -16,9 +20,13 @@ export class AppService {
     return this.ethersRpcProvider.getBalance(address);
   }
 
-  async getAllEventsFromBlockNumber(blockNumber: number, contractAddress: string): Promise<object[]> {
+  async getAllEventsFromBlockNumber(blockNumber: number): Promise<object[]> {
     const currentBlockNumber = await this.ethersRpcProvider.getBlockNumber();
-    const contract = new ethers.Contract(contractAddress, BridgeV1__factory.abi, this.ethersRpcProvider);
+    const contract = new ethers.Contract(
+      this.configService.getOrThrow('contract.address'),
+      BridgeV1__factory.abi,
+      this.ethersRpcProvider,
+    );
     const eventSignature = contract.filters.BRIDGE_TO_DEFI_CHAIN();
     const events = await contract.queryFilter(eventSignature, blockNumber, currentBlockNumber - 65);
     return events;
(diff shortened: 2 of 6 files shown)


weiyuan95:
Writing this out here so I don't forget.

Next steps that come after this PR. These can probably be separate PRs to reduce reviewer scope.
- Last-checked block number persistence
  - Block number should not need to be passed in via an API call
  - This involves any necessary DB work that needs to be done to persist the data
- Bridge proxy address should be passed in as a config variable instead of an API call

weiyuan95 (changes requested):

apps/server/src/AppConfig.ts
```diff
@@ -11,6 +11,9 @@ export function appConfig() {
     ethereum: {
       rpcUrl: process.env.ETHEREUM_RPC_URL || 'localhost:8545',
     },
+    contract: {
+      address: process.env.BRIDGE_CONTRACT_ADDRESS || '',
```
lykalabrada, line 15: If we are to set a config like this as environment variable, I wonder how it's possible to manage for different networks (testnet, mainnet,..).

This is a general question as I encountered something similar and had to create separate variables, something like  `goerli_contract_address, mainnet_contract_address` instead

I wonder how we'll handle different networks here 🤔

weiyuan95: Great point! Since the contract addresses are immutable, the testnet contract address can actually already be hardcoded to `0x93fE70235854e7c97A5db5ddfC6eAAb078e99d3C`, and similarly for the mainnet contract when that gets deployed.

In terms of how that would look like on the `config`, something like `contracts.bridgeProxy.mainnet.address` and `contracts.bridgeProxy.testnet.address` makes sense to me (or `mainnetAddress` and `testnetAddress` if we don't care about how the 'object' looks like)

apps/server/src/AppService.ts
```diff
@@ -14,4 +19,16 @@ export class AppService {
...
+
+  async getAllEventsFromBlockNumber(blockNumber: number): Promise<object[]> {
+    const currentBlockNumber = await this.ethersRpcProvider.getBlockNumber();
+    const contract = new ethers.Contract(
+      this.configService.getOrThrow('contract.address'),
+      BridgeV1__factory.abi,
+      this.ethersRpcProvider,
+    );
+    const eventSignature = contract.filters.BRIDGE_TO_DEFI_CHAIN();
+    const events = await contract.queryFilter(eventSignature, blockNumber, currentBlockNumber - 65);
```
Abhishekkochar, line 31: @greggtan,general question, is there a specific reason why we decide on checking past 65 blocks?

apps/server/src/AppService.ts
```diff
@@ -14,4 +19,16 @@ export class AppService {
   async getBalance(address: string): Promise<BigNumber> {
     return this.ethersRpcProvider.getBalance(address);
   }
+
+  async getAllEventsFromBlockNumber(blockNumber: number): Promise<object[]> {
+    const currentBlockNumber = await this.ethersRpcProvider.getBlockNumber();
+    const contract = new ethers.Contract(
+      this.configService.getOrThrow('contract.address'),
```
eli-lim, line 26: This should be sourced in the constructor instead, so if the contract address doesn't exist, then the app can fail start-up (we prefer early exits instead of silent time-bombs in production)

apps/server/src/AppConfig.ts
eli-lim, line 15: this shouldn't default to an empty string - it should either be a valid address or `undefined`

apps/server/src/AppService.ts
greggtan: Following issue 183 (https://github.com/WavesHQ/bridge/issues/183) by @weiyuan95, we will wait for 65 confirmations, unless we want to wait be speedier at the expense of some security. In that case, we will go with 33 confirmations!

This number is definitely flexible and will depend on discussion.

apps/server/test-i9n/BridgeApp.i9n.ts
```diff
@@ -1,48 +1,217 @@
...
 import { BridgeServerTestingApp } from '../src/BridgeServerTestingApp';
 
 @Module({})
 export class TestingExampleModule {
-  static register(startedHardhatContainer: StartedHardhatNetworkContainer): DynamicModule {
-    const hardhatConfig = registerAs('ethereum', () => ({
-      rpcUrl: startedHardhatContainer.rpcUrl,
-    }));
-
+  static register(config: AppConfig): DynamicModule {
```
Abhishekkochar, line 21: It would be best to have `TestingExampleModule` in separate file as it being used in different files.

apps/server/test-i9n/BridgeApp.i9n.ts
```diff
@@ -1,48 +1,217 @@
...
     return {
       module: TestingExampleModule,
-      imports: [AppModule, ConfigModule.forFeature(hardhatConfig)],
+      imports: [AppModule, ConfigModule.forFeature(() => config)],
     };
   }
 }
 
+export function buildTestConfig(startedHardhatContainer: StartedHardhatNetworkContainer, contractAddress?: string) {
```
Abhishekkochar, line 29: Same as above. We can also do strongly type return 
```suggestion
export function buildTestConfig(startedHardhatContainer: StartedHardhatNetworkContainer, contractAddress?: string) : AppConfig {....
```

Abhishekkochar (commented):
Everything else LGTM.

apps/server/src/AppService.ts
```diff
@@ -1,11 +1,24 @@
...
 export class AppService {
-  constructor(@Inject(ETHERS_RPC_PROVIDER) readonly ethersRpcProvider: ethers.providers.StaticJsonRpcProvider) {}
+  private contract: Contract;
+
+  constructor(
+    @Inject(ETHERS_RPC_PROVIDER) readonly ethersRpcProvider: ethers.providers.StaticJsonRpcProvider,
+    private configService: ConfigService,
+  ) {
+    this.contract = new ethers.Contract(
+      this.configService.getOrThrow('contract.bridgeProxy.testnetAddress'),
```
cuongquangnam, line 17: Why does it have to be a testnet address over here? Can it be a mainnet address?

weiyuan95: Clarified with Vir. The UI will support both test and main net transactions, so we'll need to poll both. Creating a service that's specifically doing that will come in another PR though.

apps/server/test-i9n/BridgeApp.i9n.ts
```diff
@@ -1,48 +1,217 @@
...
+  let bridgeUpgradeable: BridgeV1;
+  let bridgeProxy: BridgeProxy;
   let testing: BridgeServerTestingApp;
 
   beforeAll(async () => {
     startedHardhatContainer = await new HardhatNetworkContainer().start();
     hardhatNetwork = await startedHardhatContainer.ready();
-    testing = new BridgeServerTestingApp(TestingExampleModule.register(startedHardhatContainer));
+
+    evmContractManager = hardhatNetwork.contracts;
```
weiyuan95, line 57: These test fixtures (ie setting up the bridge contracts) can sit in something like a `BridgeAppTesting` class to be DRY, but that is an improvement that can come in a separate PR. Let's get the baseline out first.

apps/server/test-i9n/defichain/wallet.i9n.ts
```diff
@@ -13,7 +13,7 @@ describe('DeFiChain Wallet Integration Testing', () => {
   beforeAll(async () => {
     startedHardhatContainer = await new HardhatNetworkContainer().start();
     hardhatNetwork = await startedHardhatContainer.ready();
-    testing = new BridgeServerTestingApp(TestingExampleModule.register(startedHardhatContainer));
+    testing = new BridgeServerTestingApp(TestingExampleModule.register(buildTestConfig(startedHardhatContainer, '')));
```
weiyuan95, line 16:
```suggestion
    testing = new BridgeServerTestingApp(TestingExampleModule.register(buildTestConfig(startedHardhatContainer)));
```

Since it's optional, we don't need to pass it in. Same for the other test.

apps/server/test-i9n/BridgeApp.i9n.ts
weiyuan95: Another thing, since i'm guessing this test config will only get larger (postgres, etc), with different types, I would prefer that this is as explicit as possible with what users need to pass in. And ideal function call would be

```ts
buildTestConfig({
  hardhatContainer: startedHardhatContainer,
  bridgeProxyAddress: 'some_address',
  ...
})
```
This way it clear (something like named params) on what people need to pass in to build the test config.

weiyuan95 (changes requested):

weiyuan95 (approved):
Merging this in first as a first-cut baseline. Follow up tasks from comments [here](https://github.com/WavesHQ/bridge/issues/281)

PR #289, opened 2023-02-02
feat(Contracts): Remove supports for ETHER

### LINKED ISSUE #276: chore(contracts): Remove native ETH support
Remove native ETH support from the smart contracts. The bridge contract should now only support ERC20 wETH.

#### Why is this needed:
Consolidates the supported tokens implementation. wETH also has more widespread use, and allows users to easily participate in DeFi/smart contracts on Ethereum.

Abhishekkochar:
Removing the support of Ether token. Bridge contract will no longer be able to accept Ether anymore.  

Fixes #276 

#### Developer Checklist:

- [ ] Unit tests\*
- [ ] Added e2e tests\*
- [ ] Added translations\*

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,22 +43,14 @@ To change the state of any smart contract, users need to approve the smart contr
 
 Once approved, user will call the `bridgeToDeFiChain()` function with following arguments: `_defiAddress`- address on Defi Chain that receiving funds, `_tokenAddress` - ERC20 token's address and `_amount` amount to bridge over to Defi chain.
 
-### Fund Ether - to transfer ERC20 tokens from an EOA to the Bridge
-
-When sending ETHER to bridge, the user will not have to approve the contract. By default, every smart contract accepts ETH. Sending ether will be similar to an ERC20 token, except we don't account for `_amount` - instead `msg.value()` is used. The `_defiAddress` is the address on the DeFiChain that is receiving funds. Since ETH does not have an address since it is the native currency, it is identified by address(0)/0x0 in `_tokenAddress`
-
 ### Add supported token
 
-Only addresses with the Admin and Operational role can call the `addSupportedTokens()` function. This sets the `_dailyAllowance` for a ERC20 token identified by its `_tokenAddress`. The `_startAllowanceTimeFrom` also represents when this token 'goes live'
+Only addresses with the Admin and Operational roles can call the `addSupportedTokens()` function. This sets the `_dailyAllowance` for an ERC20 token identified by its `_tokenAddress`. The `_startAllowanceTimeFrom` also represents when this token 'goes live'
 User are not allowed to bridge more than the dailyAllowance per day.
 
 ### Remove supported token
 
 Only addresses with the Admin and Operational role can call the `removeSupportedTokens()` function.
-
-### Withdraw ether
-
-Only the Admin can call the `withdrawEth()` function with ETH's amount.
 
 ### Withdraw ERC20
 
@@ -70,9 +62,9 @@ Both the Admin and Operational addresses can change the `_dailyAllowance` (the n
 
 During this 'change in allowance' period, no bridging to DeFiChain will be allowed. However, it is still possible to make additional changes by calling `changeDailyAllowance()` in case mistakes were made.
 
-### Withdraw / withdrawEth
+### Withdraw
 
-`withdraw()` and `withdrawEth()` functions when called will withdraw ERC20 token and ETHER respectively. Only the address with the Admin role can call these functions.
+`withdraw()` function when called will withdraw an ERC20 token. Only the address with the Admin role can call this function.
 
 ### Change relayer address
 
@@ -91,7 +83,7 @@ Only address with admin role can change `transactionFee`. Initial fee will be se
 
 ### Deploy ERC20 tokens 'MUSDT' & 'MUSDC'
 
-To deploy ERC20 token user will have to run a command `npx hardhat run --network goerli ./scripts/deployERC20.ts` in smartContract directory.
+To deploy ERC20 tokens user will have to run a command `npx hardhat run --network goerli ./scripts/deployERC20.ts` in smartContract directory.
 
 To verify the said tokens and other contracts, there would be a prompt on terminal after running the deployment command that devs will need to run after.
 
--- a/packages/smartcontracts/src/contracts/BridgeV2.sol
+++ b/packages/smartcontracts/src/contracts/BridgeV2.sol
@@ -1,0 +1,370 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.13;
+import '@openzeppelin/contracts/token/ERC20/IERC20.sol';
+import '@openzeppelin/contracts/token/ERC20/extensions/IERC20Metadata.sol';
+import '@openzeppelin/contracts-upgradeable/utils/cryptography/EIP712Upgradeable.sol';
+import '@openzeppelin/contracts-upgradeable/utils/cryptography/ECDSAUpgradeable.sol';
+import '@openzeppelin/contracts-upgradeable/access/AccessControlUpgradeable.sol';
+import '@openzeppelin/contracts-upgradeable/proxy/utils/UUPSUpgradeable.sol';
+import 'hardhat/console.sol';
+
+/** @notice @dev  
+/* This error occurs when token is in change allowance period.
+*/
+error STILL_IN_CHANGE_ALLOWANCE_PERIOD();
+
+/** @notice @dev  
+/* This error occurs when incoorect nonce provided
+*/
+error INCORRECT_NONCE();
+
+/** @notice @dev  
+/* This error occurs when token is not in supported list
+*/
+error TOKEN_NOT_SUPPORTED();
+
+/** @notice @dev  
+/* This error occurs when token is not in supported list with `timeLeft` until the token is supported
+*/
+error TOKEN_NOT_SUPPORTED_YET(uint256 timeLeft);
+
+/** @notice @dev  
+/* This error occurs when fake signatures being used to claim fund
+*/
+error FAKE_SIGNATURE();
+
+/** @notice @dev  
+/* This error occurs when bridging amount exceeds dailyAllowance limit
+*/
+error EXCEEDS_DAILY_ALLOWANCE();
+
+/** @notice @dev  
+/* This error occurs when token is already in supported list
+*/
+error TOKEN_ALREADY_SUPPORTED();
+
+/** @notice @dev  
+/* This error occurs when using Zero address
+*/
+error NON_ZERO_ADDRESS();
+
+/** @notice @dev  
+/* This error occurs when the msg.sender doesn't have neither DEFAULT_ADMIN_ROLE or OPERATIONAL_ROLE assigned
+*/
+error NON_AUTHORIZED_ADDRESS();
+
+/** @notice @dev 
+/* This error occurs when Admin(s) try to change daily allowance of un-supported token.
+*/
+error ONLY_SUPPORTED_TOKENS();
+
+/** @notice @dev 
+/* This error occurs when `_newResetTimeStamp` is before block.timestamp
+*/
+error INVALID_RESET_EPOCH_TIME();
+
+/** @notice @dev 
+/* This error occurs when `_amount` is zero
+*/
+error INVALID_AMOUNT();
+
+contract BridgeV2 is UUPSUpgradeable, EIP712Upgradeable, AccessControlUpgradeable {
+    struct TokenAllowance {
+        uint256 latestResetTimestamp;
+        uint256 dailyAllowance;
+        uint256 currentDailyUsage;
+    }
+
+    // Mapping to track the address's nonce
+    mapping(address => uint256) public eoaAddressToNonce;
+
+    // Mapping to track supported token
+    mapping(address => bool) public supportedTokens;
+
+    address public relayerAddress;
+
+    bytes32 constant DATA_TYPE_HASH =
+        keccak256('CLAIM(address to,uint256 amount,uint256 nonce,uint256 deadline,address tokenAddress)');
+
+    // Mapping to track token address to TokenAllowance
+    mapping(address => TokenAllowance) public tokenAllowances;
+
+    bytes32 public constant OPERATIONAL_ROLE = keccak256('OPERATIONAL_ROLE');
+
+    // Initial Tx fee 0.3%. Based on dps (e.g 1% == 100dps)
+    uint256 public transactionFee;
+
+    /**
+     * @notice Emitted when the user claims funds from the bridge
+     * @param tokenAddress Token that is being claimed
+     * @param to Address that funds  will be transferred to
+     * @param amount Amount of the token being claimed
+     */
+    event CLAIM_FUND(address indexed tokenAddress, address indexed to, uint256 indexed amount);
+
+    /**
+     * @notice Emitted when the user bridges token to DefiChain
+     * @param defiAddress defiAddress DeFiChain address of user
+     * @param tokenAddress Supported token's being bridged
+     * @param amount Amount of the token being bridged
+     * @param timestamp TimeStamp of the transaction
+     */
+    event BRIDGE_TO_DEFI_CHAIN(
+        bytes defiAddress,
+        address indexed tokenAddress,
+        uint256 indexed amount,
+        uint256 indexed timestamp
+    );
+
+    /**
+     * @notice Emitted when a new token is being added to the supported list by only Admin accounts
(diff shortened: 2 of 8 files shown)


packages/smartcontracts/src/contracts/BridgeV2.sol
```diff
@@ -0,0 +1,371 @@
...
+        address _to,
+        uint256 _amount,
+        uint256 _nonce,
+        uint256 _deadline,
+        address _tokenAddress,
+        bytes calldata signature
+    ) external {
+        if (eoaAddressToNonce[_to] != _nonce) revert INCORRECT_NONCE();
+        if (!supportedTokens[_tokenAddress]) revert TOKEN_NOT_SUPPORTED();
+        if (block.timestamp > _deadline) revert INVALID_RESET_EPOCH_TIME();
```
cuongquangnam, line 221: Not valid error name, should be named like: EXPIRED_CLAIM

packages/smartcontracts/src/contracts/BridgeV2.sol
```diff
@@ -0,0 +1,371 @@
...
+     * @param _defiAddress DefiChain token address
+     * @param _tokenAddress Supported token address that being bridged
+     * @param _amount Amount to be bridged, this in in Wei
+     */
+    function bridgeToDeFiChain(
+        bytes calldata _defiAddress,
+        address _tokenAddress,
+        uint256 _amount
+    ) external {
+        if (_tokenAddress == address(0)) revert NON_ZERO_ADDRESS();
```
cuongquangnam, line 242: No need for this check, as supportedTokens[0x0] would be never be true

packages/smartcontracts/src/contracts/BridgeV2.sol
```diff
@@ -0,0 +1,371 @@
...
+     * @param newTxFee New transcation fee in bps
+     */
+    event TRANSACTION_FEE_CHANGED(uint256 oldTxFee, uint256 newTxFee);
+
+    /**
+     * @notice Emitted when withdrawal of ETHER only by the Admin account
+     * @param ownerAddress Owner's address requesting withdraw
+     * @param withdrawalAmount amount of respected token being withdrawn
+     */
+    event ETH_WITHDRAWAL_BY_OWNER(address ownerAddress, uint256 withdrawalAmount);
```
cuongquangnam, line 173: redundant event

packages/smartcontracts/src/contracts/BridgeV2.sol
```diff
@@ -0,0 +1,371 @@
...
+
+/** @notice @dev  
+/* This error occurs when token is not in supported list
+*/
+error TOKEN_NOT_SUPPORTED();
+
+/** @notice @dev  
+/* This error occurs when token is not in supported list with `timeLeft` until the token is supported
+*/
+error TOKEN_NOT_SUPPORTED_YET(uint256 timeLeft);
```
cuongquangnam, line 29: redundant error

packages/smartcontracts/src/contracts/BridgeV2.sol
```diff
@@ -0,0 +1,371 @@
...
+    function bridgeToDeFiChain(
+        bytes calldata _defiAddress,
+        address _tokenAddress,
+        uint256 _amount
+    ) external {
+        if (_tokenAddress == address(0)) revert NON_ZERO_ADDRESS();
+        if (!supportedTokens[_tokenAddress]) revert TOKEN_NOT_SUPPORTED();
+        uint256 tokenAllowanceStartTime = tokenAllowances[_tokenAddress].latestResetTimestamp;
+        if (block.timestamp < tokenAllowanceStartTime) revert STILL_IN_CHANGE_ALLOWANCE_PERIOD();
+        if (_amount == 0) revert INVALID_AMOUNT();
```
cuongquangnam, line 246:
```suggestion
        if (_amount == 0) revert ZERO_AMOUNT();
```

lykalabrada:
Hello @Abhishekkochar just wanna check, does this mean we won't allow users to send ETH anymore? And only allow ERC20 tokens?

Abhishekkochar:
> Hello @Abhishekkochar just wanna check, does this mean we won't allow users to send ETH anymore? And only allow ERC20 tokens?

@lykalabrada, yes we are not supporting ETH anymore. User will need to use WETH when bridging instead.

packages/smartcontracts/src/contracts/BridgeV2.sol
Abhishekkochar: Yes, you are right. But admins can still add zero address as supported token. This is more of a sanity check.

lykalabrada:
> @lykalabrada, yes we are not supporting ETH anymore. User will need to use WETH when bridging instead.

I see. Do you mind deploying one more test token for Mock wETH, then add the token address in README? 🙏

Abhishekkochar:
> > @lykalabrada, yes we are not supporting ETH anymore. User will need to use WETH when bridging instead.
> 
> I see. Do you mind deploying one more test token for Mock wETH, then add the token address in README? 🙏

WETH is already deployed here: https://goerli.etherscan.io/address/0xb4fbf271143f4fbf7b91a5ded31805e42b2208d6#writeContract

You need to call the deposit() and input desired ETH need to wrapped.

packages/smartcontracts/src/contracts/BridgeV2.sol
cuongquangnam: This is bridgeToDeFiChain function. We alr added the sanity check in addSupportedTokens, therefore 0x0 will never be in the list of supportedTokens --> check of supportedTokens[_tokenAddress] in the line below will eliminate the case for 0x0.

Abhishekkochar: Yes, you are right. I mistook it for `addSupportedTokens` 😅😅

weiyuan95:
> > > @lykalabrada, yes we are not supporting ETH anymore. User will need to use WETH when bridging instead.
> > 
> > 
> > I see. Do you mind deploying one more test token for Mock wETH, then add the token address in README? 🙏
> 
> WETH is already deployed here: https://goerli.etherscan.io/address/0xb4fbf271143f4fbf7b91a5ded31805e42b2208d6#writeContract
> 
> You need to call the deposit() and input desired ETH need to wrapped.

Can we include this in the README as well just for convenience's sake? 😄

packages/smartcontracts/src/tests/BridgeProxy.supportedToken.spec.ts
```diff
@@ -6,111 +6,117 @@ import { deployContracts } from './testUtils/deployment';
...
-        await proxyBridge
+  describe('DEFAULT_ADMIN_ROLE', () => {
+    it('Successfully add token to supported list & allowance by Admin role address', async () => {
+      const { proxyBridge, testToken, testToken2, defaultAdminSigner } = await loadFixture(deployContracts);
+      await proxyBridge.addSupportedTokens(testToken.address, toWei('15'), getCurrentTimeStamp());
+      // Adding the testToken2 as the supported token by Admin role only.
+      expect(await proxyBridge.supportedTokens(testToken2.address)).to.equal(false);
+      await proxyBridge
+        .connect(defaultAdminSigner)
+        .addSupportedTokens(testToken2.address, toWei('15'), getCurrentTimeStamp());
```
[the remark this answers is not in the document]
weiyuan95: Hmm. I would say it's fine because
- we will never set a 0 timestamp, it will already be checked in the multi sig transaction
- even if we do set a 0 timestamp, the timestamp will update on the next day to be the right timestamp

Unless i'm missing something? 🤔

packages/smartcontracts/src/tests/BridgeProxy.defiToEvm.spec.ts
```diff
@@ -143,36 +143,61 @@ describe('DeFiChain --> EVM', () => {
...
+        getCurrentTimeStamp() - 60 * 60 * 1,
+        testToken.address,
+        signature,
+      ),
+    ).to.be.revertedWithCustomError(proxyBridge, 'EXPIRED_CLAIM');
+    // Checking Balance after claiming fund, should be 0
+    expect(await testToken.balanceOf(defaultAdminSigner.address)).to.equal(toWei('0'));
+  });
+
+  it('Successfully revert when claim more than available amount', async () => {
```
weiyuan95, line 173: > more than the available amount

Available amount/balance on the bridge? Is this to catch the scenario where the funds on the bridge are not maintained?

Abhishekkochar: Testing the case where the contract's balance is less than the claim's balance

packages/smartcontracts/src/tests/BridgeProxy.supportedToken.spec.ts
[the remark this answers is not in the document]
Abhishekkochar: @weiyuan95, ran tests for this specific case. `latestResetTimestamp` will set to the GMT+8 (Mon Feb 06 2023 08:00:00 GMT+0800) if today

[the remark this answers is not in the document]
cuongquangnam: alright, I just feel it is more natural if we can add at least a sanity check in addSupportedTokens that block.timestamp < _startAllowanceTimeFrom. Otherwise, it is ok to me!

README.md
```diff
@@ -48,6 +48,8 @@ Once approved, user will call the `bridgeToDeFiChain()` function with following
 Only addresses with the Admin and Operational roles can call the `addSupportedTokens()` function. This sets the `_dailyAllowance` for an ERC20 token identified by its `_tokenAddress`. The `_startAllowanceTimeFrom` also represents when this token 'goes live'
 User are not allowed to bridge more than the dailyAllowance per day.
 
+If the admin sets `_startAllowanceTimeFrom` to 00:00 when adding a supported token, the token will be immediately supported. Upon bridging the token for the first time, `latestResetTimestamp` will be reset to 08:00:00 SGT (GMT+0800) for the day.
```
weiyuan95, line 51:
```suggestion
If the admin sets `_startAllowanceTimeFrom` to 0 when adding a supported token, the token will be immediately supported. Upon bridging the token for the first time, `latestResetTimestamp` will be reset to 08:00:00 SGT (GMT+0800) for the day.
```

I think it's just 0 here and no 00:00? 🤔

packages/smartcontracts/src/tests/BridgeProxy.supportedToken.spec.ts
```diff
@@ -153,15 +153,29 @@ describe('Add and Removed Supported ETH and ERC20 tokens', () => {
...
+    expect((await proxyBridge.tokenAllowances(testToken.address)).dailyAllowance).to.equal(toWei('15'));
     // Bridging 10 test tokens
     await proxyBridge.bridgeToDeFiChain(ethers.constants.AddressZero, testToken.address, toWei('10'));
+    // Getting 08:00:00 GMT+0800 (SGT) for the day
+    const date = new Date();
+    date.setUTCHours(0);
+    date.setUTCMinutes(0);
+    date.setUTCSeconds(0);
+    date.setUTCMilliseconds(0);
+    const sgtTimeZoneSeconds = date.getTime() / 1000;
```
weiyuan95, line 175:
```suggestion
    // The unix timestamp at 00:00 GMT
    const utcMidnightTimestamp = date.getTime() / 1000;
```

Would use this instead of `sgtTimeZone` since the timestamp here is `    // The unix timestamp at 00:00 GMT`

weiyuan95 (changes requested):

packages/smartcontracts/src/tests/BridgeProxy.supportedToken.spec.ts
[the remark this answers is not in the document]
Abhishekkochar: Yes, You are right. We have kinda similar check in `changeDailyAllowance`. We shouldn't be able to add past time stamp anyway. @weiyuan95, your thoughts on this please. Also, We wouldn't need the below test anymore except the check for this.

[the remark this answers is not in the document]
weiyuan95: Let's do it that way then.

I don't have much of an opinion on this, but it will definitely be slightly more irritating if we want to have the concept of 'instant support'. I don't really see that happening much though, so let's have that sanity check in.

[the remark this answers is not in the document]
Abhishekkochar: As for instant support, we can do something like this `getCurrentTimeStamp({ additionalTime: 60 })`  Supporting in a min would be enough to meet this requirement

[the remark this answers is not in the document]
weiyuan95: That will be slightly more irritating, since we need to get the current timestamp + add some minutes to try and guess the block timestamp that it gets mined in.

Regardless, I highly doubt we will have that use case - and even if we did, it will not need to be __instantly__ supported. Let's implement the sanity check.

weiyuan95 (dismissed):
LGTM! Will see if @cuongquangnam has any other comments. Let's get this in by tomorrow.

Abhishekkochar:
@weiyuan95 & @cuongquangnam , to avoid any confusion I have removed implementation to BridgeV1 smart contract. As this is only deployed to testnet, we don't need to keep both of these contracts.

weiyuan95:
I think we need to fix other tests due to the contract changes

Abhishekkochar:
> I think we need to fix other tests due to the contract changes

Screenshot 2023-02-08 at 10 50 23 AM

Issue seems to be in sever/ integration app tests

packages/smartcontracts/src/contracts/BridgeV1.sol
```diff
@@ -410,28 +353,12 @@ contract BridgeV1 is UUPSUpgradeable, EIP712Upgradeable, AccessControlUpgradeabl
...
-        } else return _amount;
-    }
-
     /**
      * This function provides the net amount after deducting fee
      * @param _amount Ideally will be the value of erc20 token
      * @return netAmountInWei net balance after the fee amount taken
      */
-    function amountAfterFees(uint256 _amount) internal view returns (uint256 netAmountInWei) {
+    function amountAfterFees(uint256 _amount) private view returns (uint256 netAmountInWei) {
```
weiyuan95, line 361: Hmm, weren't we changing this to `internal` instead? Can't remember the reason.

packages/smartcontracts/src/contracts/BridgeV1.sol
```diff
@@ -121,10 +110,10 @@ contract BridgeV1 is UUPSUpgradeable, EIP712Upgradeable, AccessControlUpgradeabl
      * @param timestamp TimeStamp of the transaction
      */
     event BRIDGE_TO_DEFI_CHAIN(
-        bytes indexed defiAddress,
+        bytes defiAddress,
```
weiyuan95, line 113: I think we were planning to index the `defiAddress` instead?

packages/smartcontracts/src/tests/BridgeProxy.evmToDefi.spec.ts
```diff
@@ -13,15 +13,15 @@ async function initMintAndSupport(
   testToken: TestToken,
   eoaAddress: string,
   contractAddress: string,
-  additionalTime?: number,
+  additionalTime: number,
```
weiyuan95, line 16: Hmm, does this actually need to be a parameter? Can we not just arbitrarily pass in the `additionalTime`? Ie instead of a param it's just `getCurrentTimeStamp({additionalTime: 15})`

From what I see you seem to be always passing in `15` 🤔

weiyuan95 (changes requested):
Some quick comments. Can we quickly do a quick scan of the v1 contract as well? There might have been some unintended changes when merging v1 and v2

packages/smartcontracts/src/tests/BridgeProxy.evmToDefi.spec.ts
Abhishekkochar: Supporting 15secs ahead in time for the instant support. Some places i'm using different value.

Abhishekkochar: In some cases passing 15 secs for instant support. On other it's 24hrs.

weiyuan95: Ah ok, as long as we are actually using this parameter.

weiyuan95 (approved):
LGTM - merging by EOD if @cuongquangnam has no comments.

PR #297, opened 2023-02-03
feature(server): DeFiChain Transaction service

### LINKED ISSUE #192: DeFiChain Transaction builder
To be used for `minttoken` and `burntoken`. Allows the server to craft and broadcast DeFiChain transactions. - https://github.com/WavesHQ/bridge/issues/185

#### Why is this needed:
This transaction builder will accept an account (WhaleWalletAccount) that will be used for signing transactions.

thedoublejay:
1. Setup `DeFiChainStubContainer` - Allows to run Whale, Playground and Node for easier testing. Thanks to @fuxingloh for the assist. 🚀
2. With the setup of local DeFiChain Network, remove the reliance to "Remote Playground" for more consistent testing.
3. TransactionService - to craft/broadcast DeFiChain txs
4. Send service - Crafts a send tx via TransactionService. Supports sending of UTXO (DFI) or other dTokens (account to account)

Fixes #192 

#### Additional comments?:
- [ ] Managing version bumps on docker pulls (to make tests faster) https://github.com/WavesHQ/bridge/issues/316

#### Developer Checklist:

- [ ] Unit tests\*
- [ ] Added e2e tests\*
- [ ] Added translations\*


apps/server/test-i9n/defichain/DeFiChainStubContainer.ts
```diff
@@ -0,0 +1,57 @@
...
+    this.whale = await new WhaleApiContainer().withNetwork(network).withNativeChain(this.defid, network).start();
+    this.playground = await new PlaygroundApiContainer()
+      .withNetwork(network)
+      .withNativeChain(this.defid, network)
+      .start();
+    await this.playground.waitForReady();
+    this.playgroundClient = new PlaygroundApiClient({ url: this.playground.getPlaygroundApiClientOptions().url });
+    this.playgroundRpcClient = new PlaygroundRpcClient(this.playgroundClient);
+    this.whaleClient = new WhaleApiClient(this.whale.getWhaleApiClientOptions());
+    return this.whale.getWhaleApiClientOptions().url;
```
weiyuan95, line 45: This is not ideal, since usually the mental model of a `start` method is that it either returns `void` or something related to the `start`. It's not sustainable as well. What happens if we need to return the playground url, or the defid node url?

We can do something like a `StartedDeFiChainStubContainer` that is a collection of the started containers. In that way, there is the guarantee of the whale url without needing to check if it's undefined or not.

Something like this:

```ts
class DeFiChainStubContainer {
  // no constructor
  async start(): Promise<StartedDeFiChainContainer> {
    // start containers
    return new StartedDeFiChainContainer(...startedContainers)
  }
}

class StartedDeFiChainContainer  {
  constructor(...startedContainers) {
  }

  get whaleUrl: string {
    return this.whale.url
  }
}
```
You can look at more references on the fluent (and `Started*Container` design) [here](https://github.com/BirthdayResearch/sticky/blob/main/packages/sticky-testcontainers/src/PostgreSqlContainer.ts)

apps/server/src/AppConfig.ts
```diff
@@ -5,11 +5,9 @@ export function appConfig() {
   return {
     defichain: {
       [EnvironmentNetwork.MainNet]: process.env.DEFICHAIN_MAINNET_KEY,
-      [EnvironmentNetwork.RemotePlayground]:
-        process.env.DEFICHAIN_REGTEST_KEY ||
-        'avoid between cupboard there nerve sugar quote foot broom intact seminar culture much anger hold rival moral silly volcano fog service decline tortoise combine',
-      [EnvironmentNetwork.LocalPlayground]:
-        'avoid between cupboard there nerve sugar quote foot broom intact seminar culture much anger hold rival moral silly volcano fog service decline tortoise combine',
+      [EnvironmentNetwork.RemotePlayground]: process.env.DEFICHAIN_REMOTE_KEY,
```
weiyuan95, line 8: Personally not a big fan of using dynamic variables for configs, these should be hardcoded anyway and should not change.

apps/server/test-i9n/defichain/send.i9n.ts
```diff
@@ -0,0 +1,104 @@
...
+  let wallet: WhaleWalletAccount;
+  const toAddress = 'bcrt1q8rfsfny80jx78cmk4rsa069e2ckp6rn83u6ut9';
+
+  beforeAll(async () => {
+    defichain = await new DeFiChainStubContainer();
+    const localWhaleURL = await defichain.start();
+    const dynamicModule = TestingModule.register(
+      buildTestConfig({ defichain: { localWhaleURL, localDefichainKey: defichain.localMnemonic } }),
+    );
+    dynamicModule.providers = [
```
weiyuan95, line 30: What's the need to override the providers here? 🤔

apps/server/test-i9n/defichain/send.i9n.ts
```diff
@@ -0,0 +1,104 @@
...
+    // Check balance if was sent properly
+    const response = await defichain.whaleClient?.address.listToken(toAddress);
+    expect(response?.[0].id).toStrictEqual(token.id);
+    expect(response?.[0].amount).toStrictEqual(token.amount.toFixed(8));
+    expect(response?.[0].symbol).toStrictEqual(token.symbol);
+  });
+
+  it('should be able to send DFI (UTXO)', async () => {
+    // Tests are slower because it's running 3 containers at the same time
+    jest.setTimeout(3600000);
```
weiyuan95, line 87: If necessary these can be set on the entire file (top level) or as a global config instead of individually in the tests

weiyuan95 (changes requested):

apps/server/src/defichain/services/SendService.ts
```diff
@@ -0,0 +1,46 @@
...
+import { EnvironmentNetwork } from '@waveshq/walletkit-core';
+import BigNumber from 'bignumber.js';
+
+import { TransactionService } from './TransactionService';
+
+@Injectable()
+export class SendService {
+  constructor(private readonly transactionService: TransactionService) {}
+
+  async send(
```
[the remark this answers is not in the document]
thedoublejay: My idea with transaction service is it will be used for all other transactions. It's separated with services (Send). If in the future we want to have another transaction (e.g, PoolSwap, Convert etc.), we will just have another service for it and not lump everything with TransactionService.

apps/server/src/defichain/services/SendService.ts
```diff
@@ -0,0 +1,46 @@
...
+        // Rest of dTokens to use this tx type
+        signed = await builder.account.accountToAccount(
+          {
+            from,
+            to: [
+              {
+                script: to,
+                balances: [
+                  {
+                    token: +token.id,
```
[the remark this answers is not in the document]
thedoublejay: I remember this was an issue on node side to ensure that it's always a numbered id

apps/server/src/defichain/services/TransactionService.ts
```diff
@@ -0,0 +1,91 @@
...
+      // Known issue on DeFiChain, need to add retry on broadcast
+      if (retries < 3) {
+        return await this.broadcastTransaction(network, tx, retries + 1);
+      }
+      throw e;
+    }
+  }
+
+  // Check if the transaction has been confirmed
+  async waitForTxConfirmation(network: EnvironmentNetwork, id: string): Promise<Transaction> {
```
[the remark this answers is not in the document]
thedoublejay: Yep, will be used on the DFC -> EVM process. I will create a test for it on another PR once that's implemented

apps/server/src/AppConfig.ts
thedoublejay: This will be how we inject the mnemonic keys on deployment. It will always be dynamic

apps/server/test-i9n/defichain/send.i9n.ts
thedoublejay: I was following the services testing guide haha let me see if we can remove it

apps/server/test-i9n/defichain/send.i9n.ts
thedoublejay: actually it doesn't work. I just wanted to test it out but it seems not working

apps/server/src/defichain/services/TransactionService.ts
```diff
@@ -0,0 +1,91 @@
...
+
+import { WhaleApiClientProvider } from '../providers/WhaleApiClientProvider';
+import { WhaleWalletProvider } from '../providers/WhaleWalletProvider';
+import { WhaleApiService } from './WhaleApiService';
+
+const MAX_TIMEOUT = 300000;
+const INTERVAL_TIME = 5000;
+
+@Injectable()
+export class TransactionService {
```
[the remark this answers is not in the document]
thedoublejay: updated!

apps/server/test-i9n/defichain/DeFiChainStubContainer.ts
thedoublejay: I'm putting this on another PR

lykalabrada (approved):

pierregee (approved):

PR #310, opened 2023-02-06
feat(PostgreSQL & EVMTransactionConfirmer service rework): Created a Prisma Service that is made available to be injected globally, and reworked EVMTransactionConfirmer service logic to use this PostgreSQL service

greggtan:

1. Created a with-db bash script to take care of spinning up a local db container and waiting for it to be ready prior to any prisma command that requires a database connection.

To spin up the docker container before running tests:
- In terminal (/bridge), run `cd apps/server`
- Run `./with-db generate`
- Run `./with-db migrate dev`
- Run`pnpm run test` to run reworked integration test

2. Created a Prisma Service that is made available to be injected globally

3. Reworked EVMTransactionConfirmer service logic to use this PostgreSQL service to track the block number of the last checked block. It uses the local PostgreSQL database record's block number to check for events emitted from **last checked block number** to **current block number - 65**

- If current block number -64 > last checked block number, store **last checked block number = current block number -64**

281

Fixes #

#### Developer Checklist:

- [ ] Unit tests\*
- [x] Added e2e tests\*

### DIFF
--- a/apps/server/src/AppConfig.ts
+++ b/apps/server/src/AppConfig.ts
@@ -1,7 +1,10 @@
 import * as Joi from 'joi';
 
+export const DATABASE_URL = 'DATABASE_URL';
+
 export function appConfig() {
   return {
+    [DATABASE_URL]: process.env[DATABASE_URL],
     defichain: {
       mainnet: process.env.DEFICHAIN_MAINNET_KEY,
       regtest:
@@ -31,4 +34,5 @@ export const ENV_VALIDATION_SCHEMA = Joi.object({
   ETHEREUM_RPC_URL: Joi.string().ip(),
   DEFICHAIN_MAINNET_KEY: Joi.string(),
   DEFICHAIN_REGTEST_KEY: Joi.string(),
+  [DATABASE_URL]: Joi.string().required(),
 });
--- a/apps/server/src/AppConfig.ts
+++ b/apps/server/src/AppConfig.ts
@@ -12,10 +12,28 @@ export function appConfig() {
       rpcUrl: process.env.ETHEREUM_RPC_URL || 'localhost:8545',
     },
     contract: {
-      address: process.env.BRIDGE_CONTRACT_ADDRESS || '',
+      bridgeProxy: {
+        mainnetAddress: undefined,
+        testnetAddress: '0x93fE70235854e7c97A5db5ddfC6eAAb078e99d3C',
+      },
     },
   };
 }
+export type AppConfig = {
+  defichain?: {
+    mainnet?: string;
+    regtest?: string;
+  };
+  ethereum?: {
+    rpcUrl?: string;
+  };
+  contract?: {
+    bridgeProxy?: {
+      mainnetAddress?: string;
+      testnetAddress?: string;
+    };
+  };
+};
 
 export const ENV_VALIDATION_SCHEMA = Joi.object({
   ETHEREUM_RPC_URL: Joi.string().ip(),
--- a/apps/server/src/AppConfig.ts
+++ b/apps/server/src/AppConfig.ts
@@ -11,6 +11,9 @@ export function appConfig() {
     ethereum: {
       rpcUrl: process.env.ETHEREUM_RPC_URL || 'localhost:8545',
     },
+    contract: {
+      address: process.env.BRIDGE_CONTRACT_ADDRESS || '',
+    },
   };
 }
 
--- a/apps/server/src/AppConfig.ts
+++ b/apps/server/src/AppConfig.ts
@@ -20,9 +20,11 @@ export function appConfig() {
   };
 }
 
-type DeepPartial<T> = {
-  [K in keyof T]?: T[K] extends object ? DeepPartial<T[K]> : T[K];
-};
+type DeepPartial<T> = T extends object
+  ? {
+      [P in keyof T]?: DeepPartial<T[P]>;
+    }
+  : T;
 export type AppConfig = DeepPartial<ReturnType<typeof appConfig>>;
 
 export const ENV_VALIDATION_SCHEMA = Joi.object({
--- a/apps/server/src/AppConfig.ts
+++ b/apps/server/src/AppConfig.ts
@@ -19,21 +19,11 @@ export function appConfig() {
     },
   };
 }
-export type AppConfig = {
-  defichain?: {
-    mainnet?: string;
-    regtest?: string;
-  };
-  ethereum?: {
-    rpcUrl?: string;
-  };
-  contract?: {
-    bridgeProxy?: {
-      mainnetAddress?: string;
-      testnetAddress?: string;
-    };
-  };
+
+type DeepPartial<T> = {
+  [K in keyof T]?: T[K] extends object ? DeepPartial<T[K]> : T[K];
 };
+export type AppConfig = DeepPartial<ReturnType<typeof appConfig>>;
 
 export const ENV_VALIDATION_SCHEMA = Joi.object({
   ETHEREUM_RPC_URL: Joi.string().ip(),
--- a/apps/server/src/AppController.ts
+++ b/apps/server/src/AppController.ts
@@ -1,4 +1,4 @@
-import { Controller, Get, Query } from '@nestjs/common';
+import { Body, Controller, Get, ParseIntPipe,Post, Query } from '@nestjs/common';
 import { BigNumber } from 'ethers';
 
 import { AppService } from './AppService';
@@ -16,4 +16,16 @@ export class AppController {
   async getBalance(@Query('address') address: string): Promise<BigNumber> {
     return this.appService.getBalance(address);
   }
+
+  @Post('getAllEventsFromBlockNumber')
+  async getAllEventsFromBlockNumber(
+    @Body('blockNumber', ParseIntPipe) blockNumber: number,
+    @Body('contractAddress') contractAddress: string,
+  ): Promise<object[]> {
+    return this.appService.getAllEventsFromBlockNumber(blockNumber, contractAddress);
+  }
+  // @Get('getAllEventsFromBlockNumber')
+  // async getAllEventsFromBlockNumber(@Query('blockNumber') blockNumber:number): Promise<object[]>{
+  //   return this.appService.getAllEventsFromBlockNumber(Number(blockNumber))
+  // }
 }
--- a/apps/server/src/AppController.ts
+++ b/apps/server/src/AppController.ts
@@ -1,4 +1,5 @@
-import { Controller, Get, Query } from '@nestjs/common';
+import { Body, Controller, Delete, Get, Post, Query } from '@nestjs/common';
+import { Prisma } from '@prisma/client';
 import { BigNumber, Event } from 'ethers';
 
 import { AppService } from './AppService';
@@ -18,7 +19,17 @@ export class AppController {
   }
 
   @Get('getAllEventsFromBlockNumber')
-  async getAllEventsFromBlockNumber(@Query('blockNumber') blockNumber: number): Promise<Event[]> {
-    return this.appService.getAllEventsFromBlockNumber(Number(blockNumber));
+  async getAllEventsFromBlockNumber(): Promise<Event[]> {
+    return this.appService.getAllEventsFromBlockNumber();
+  }
+
+  @Post('initDatabase')
+  async initDatabase(@Body() body: Prisma.blockNumberCreateManyInput): Promise<Prisma.BatchPayload> {
+    return this.appService.initDatabase(body);
+  }
+
+  @Delete('initDatabase')
+  async deleteDatabase(): Promise<Prisma.BatchPayload> {
+    return this.appService.deleteDatabase();
   }
 }
--- a/apps/server/src/AppController.ts
+++ b/apps/server/src/AppController.ts
@@ -1,5 +1,5 @@
 import { Controller, Get, Query } from '@nestjs/common';
-import { BigNumber } from 'ethers';
+import { BigNumber, Event } from 'ethers';
 
 import { AppService } from './AppService';
 
@@ -18,7 +18,7 @@ export class AppController {
   }
 
   @Get('getAllEventsFromBlockNumber')
-  async getAllEventsFromBlockNumber(@Query('blockNumber') blockNumber: number): Promise<object[]> {
+  async getAllEventsFromBlockNumber(@Query('blockNumber') blockNumber: number): Promise<Event[]> {
     return this.appService.getAllEventsFromBlockNumber(Number(blockNumber));
   }
 }
--- a/apps/server/src/AppController.ts
+++ b/apps/server/src/AppController.ts
@@ -1,4 +1,4 @@
-import { Body, Controller, Get, ParseIntPipe,Post, Query } from '@nestjs/common';
+import { Controller, Get, Query } from '@nestjs/common';
 import { BigNumber } from 'ethers';
 
 import { AppService } from './AppService';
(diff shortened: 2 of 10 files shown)


weiyuan95:
Need to resolve conflicts @greggtan

weiyuan95:
Prisma errors, possibly migration issues? 🤔

apps/server/src/AppService.ts
```diff
@@ -28,9 +31,39 @@ export class AppService {
...
     const currentBlockNumber = await this.ethersRpcProvider.getBlockNumber();
     const eventSignature = this.contract.filters.BRIDGE_TO_DEFI_CHAIN();
-    return this.contract.queryFilter(eventSignature, blockNumber, currentBlockNumber - 65);
+    const lastCheckedBlockNumber = await this.prisma.blockNumber.findFirst({
+      where: {
+        network: 'testnet',
+      },
+    });
+    // update last checked block number to currentBlockNumber - 64 if currentBlockNumber-64 > blockNumber in database
+    if (lastCheckedBlockNumber && currentBlockNumber - 64 > Number(lastCheckedBlockNumber.blockNumber)) {
```
weiyuan95, line 43: `lastCheckedBlockNumber` should never be undefined here, and if it is, we should probably throw an error 🤔

weiyuan95: Also, I'm thinking if this is a better way to check since it's more intuitive (bottom up instead of top down)

- If `lastCheckedBlockNumber + 64 > currentBlockNumber`
  - Exit early, there's nothing to check since there are not enough finalised blocks
- `lastCheckedBlockNumber + 64 < currentBlockNumber` (else)
  - There are 64 blocks between last checked and current, we `query` for those events)
  - We also update the `lastCheckedBlockNumber` to `lastCheckedBlockNumber + 64`

WDYT?

weiyuan95 (changes requested):

apps/server/test-i9n/BridgeApp.i9n.ts
```diff
@@ -110,10 +109,127 @@ describe('Bridge Service Integration Tests', () => {
...
+    // teardown database
+    await testing.inject({
+      method: 'DELETE',
+      url: '/app/initDatabase',
+    });
+  });
+
+  it('Returns an array of confirmed events from a given block number', async () => {
+    // Step 1: starting block should be 1003 (after initializations)
+    let currBlock = await testing.inject({
```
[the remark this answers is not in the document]
greggtan: Ah I see, was just building off the previous scaffolding which had the api call!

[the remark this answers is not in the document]
greggtan: Will make changes to call ethers provider straight

apps/server/with-db
```diff
@@ -0,0 +1,44 @@
+#!/bin/sh
```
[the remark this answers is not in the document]
greggtan: I will include a write up in the bash script, do you want me to add it to the main readme too?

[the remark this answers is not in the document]
greggtan: should I write it in the main README.md too?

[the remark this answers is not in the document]
weiyuan95: Can write in `server/README.md`

apps/server/test-i9n/BridgeApp.i9n.ts
```diff
@@ -110,10 +109,127 @@ describe('Bridge Service Integration Tests', () => {
...
     await hardhatNetwork.stop();
+
+    // teardown database
+    await testing.inject({
+      method: 'DELETE',
+      url: '/app/initDatabase',
+    });
+  });
+
+  it('Returns an array of confirmed events from a given block number', async () => {
```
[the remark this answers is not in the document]
weiyuan95: It's already merged

apps/server/test-i9n/BridgeApp.i9n.ts
```diff
@@ -163,18 +158,16 @@ describe('Bridge Service Integration Tests', () => {
...
+    expect(await hardhatNetwork.ethersRpcProvider.getBlockNumber()).toStrictEqual(1005);
 
     // step 5: calling my service should return a empty array (because it is not confirmed, blockNumber in db remains as 1003)
     let eventsArray = await testing.inject({
       method: 'GET',
       url: `/app/getAllEventsFromBlockNumber`,
     });
     await expect(JSON.parse(eventsArray.body)).toHaveLength(0);
+    let dbBlocknumber = await prismaService.blockNumber.findFirst({ where: { network: 'testnet' } });
+    expect(dbBlocknumber?.blockNumber).toStrictEqual(1003n);
```
weiyuan95, line 170: Never seen a BigInt literal before 🤣

apps/server/test-i9n/BridgeApp.i9n.ts
```diff
@@ -109,28 +111,42 @@ describe('Bridge Service Integration Tests', () => {
...
-    await testing.start();
+    const app = await testing.start();
+
+    // init postgres database
+    prismaService = app.get<PrismaService>(PrismaService);
+    await prismaService.blockNumber.createMany({
+      data: [
+        {
+          blockNumber: 1003,
+          network: 'testnet',
```
weiyuan95, line 122: Just as an FYI - we'll probably be removing the concept of test+main net on the same server. Instead, each server (or cluster of servers) will be for a particular network.

This is a change that can come later though, no need to worry about it in this PR.

apps/server/prisma/migrations/20230203090049_init/migration.sql
```diff
@@ -0,0 +1,25 @@
...
+    "index" BIGINT NOT NULL,
+    "address" TEXT NOT NULL,
+    "status" "Status" NOT NULL DEFAULT 'PENDING',
+    "createdAt" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
+
+    CONSTRAINT "PathIndex_pkey" PRIMARY KEY ("id")
+);
+
+-- CreateTable
+CREATE TABLE "blockNumber" (
```
thedoublejay, line 16: Do we want to have a standard casing for our tables?

thedoublejay (changes requested):

apps/server/src/AppService.ts
greggtan: sounds good, lets do it

weiyuan95:
Given the recent implementation considerations, let's strip this PR to just the `Prisma` related changes.

apps/server/with-db
```diff
@@ -0,0 +1,46 @@
...
+P='\033[1;35m' # Highlight - light purple
+N='\033[0m' # No color
+log() {
+  echo "${C}[$(date +"%F %T %Z")] ${P}$1${N}"
+}
+
+log "Spinning up local database..."
+
+# Start local dev db container with dev credentials
+POSTGRES_IMAGE=postgres:14.5
```
fullstackninja864, line 18: Can we use alpine version here? I guess we don't need full version

apps/server/test-i9n/BridgeApp.i9n.ts
fullstackninja864: Can convert this to number for checking response

weiyuan95: We'll be removing these entities since the implementation has changed (we're not using this implementation anymore), so no worries about this.

apps/server/prisma/schema.prisma
```diff
@@ -21,8 +26,9 @@ model PathIndex {
...
 
-model BlockNumber {
-  network     String   @id
-  blockNumber BigInt   @default(1000)
-  createdAt   DateTime @default(now())
+model BridgeEventTransactions {
+  id              BigInt            @id @default(autoincrement())
+  transactionHash String
+  status          TransactionStatus @default(NOT_CONFIRMED)
+  createdAt       DateTime          @default(now())
```
weiyuan95, line 33: Can we have `updatedAt` as well? Since we will be modifying the `status` field.

apps/server/src/AppController.ts
```diff
@@ -17,8 +17,8 @@ export class AppController {
     return this.appService.getBalance(address);
   }
 
-  @Get('getAllEventsFromBlockNumber')
-  async getAllEventsFromBlockNumber(): Promise<Event[]> {
-    return this.appService.getAllEventsFromBlockNumber();
+  @Get('checkTransactionConfirmationStatus')
+  async checkTransactionConfirmationStatus(@Query('transactionHash') transactionHash: string): Promise<boolean> {
+    return this.appService.checkTransactionConfirmationStatus(transactionHash);
```
weiyuan95, line 22:
```suggestion
  @Post('handleTransaction')
  async handleTransaction(@Body?('transactionHash') transactionHash: string): Promise<boolean> {
    return this.appService.handleTransaction(transactionHash);
```
Should be a `POST`, since this can also trigger a 'write' of data.

Down the line we need to properly separate concerns instead of lumping everything in `AppService`

apps/server/src/AppService.ts
```diff
@@ -30,36 +30,53 @@ export class AppService {
     return this.ethersRpcProvider.getBalance(address);
   }
 
-  async getAllEventsFromBlockNumber(): Promise<Event[]> {
+  async checkTransactionConfirmationStatus(transactionHash: string): Promise<boolean> {
+    const receipt = await this.ethersRpcProvider.getTransactionReceipt(transactionHash);
+    const confirmed = receipt.status === 1;
+    if (!confirmed) {
```
weiyuan95, line 36:
```suggestion
    const isReverted = receipt.status === 0;
    if (isReverted === true) {
```

I would prefer explicit equality checks, but that's just a personal nit of mine.

apps/server/src/AppService.ts
```diff
@@ -30,36 +30,53 @@ export class AppService {
     return this.ethersRpcProvider.getBalance(address);
   }
 
-  async getAllEventsFromBlockNumber(): Promise<Event[]> {
+  async checkTransactionConfirmationStatus(transactionHash: string): Promise<boolean> {
+    const receipt = await this.ethersRpcProvider.getTransactionReceipt(transactionHash);
```
weiyuan95, line 34:
```suggestion
    const txReceipt = await this.ethersRpcProvider.getTransactionReceipt(transactionHash);
```

apps/server/with-db
```diff
@@ -15,7 +15,7 @@ log() {
 log "Spinning up local database..."
 
 # Start local dev db container with dev credentials
-POSTGRES_IMAGE=postgres:14.5
+POSTGRES_IMAGE=postgres:14.6-alpine
```
weiyuan95, line 18: I would revert this to keep it at parity with https://github.com/BirthdayResearch/sticky/blob/main/packages/sticky-testcontainers/src/PostgreSqlContainer.ts#L14

The difference isn't that big anyway 🤔

apps/server/prisma/schema.prisma
```diff
@@ -13,10 +13,22 @@ enum Status {
   COMPLETED
 }
 
+enum TransactionStatus {
```
weiyuan95, line 16:
```suggestion
enum EthereumTransactionStatus {
```
Should be explicit to differentiate between ETH and DFC 🤔

apps/server/prisma/schema.prisma
```diff
@@ -13,10 +13,22 @@ enum Status {
...
   index     BigInt   @unique
   address   String
   status    Status   @default(PENDING)
   createdAt DateTime @default(now())
 }
+
+model BridgeEventTransactions {
+  id              BigInt            @id @default(autoincrement())
+  transactionHash String
+  status          TransactionStatus @default(NOT_CONFIRMED)
```
weiyuan95, line 32:
```suggestion
  status          TransactionStatus
```
I would say not to set a default for this to force users to specify the status.

apps/server/src/AppService.ts
```diff
@@ -28,9 +30,53 @@ export class AppService {
...
-    return this.contract.queryFilter(eventSignature, blockNumber, currentBlockNumber - 65);
+    const numberOfConfirmations = currentBlockNumber - receipt.blockNumber;
+
+    const txHashFound = await this.prisma.bridgeEventTransactions.findFirst({
+      where: {
+        transactionHash,
+      },
+    });
+
+    if (!txHashFound && numberOfConfirmations < 65) {
```
weiyuan95, line 48: Would prefer nested conditionals for this, since it 'self documents' the code
```ts
// New transaction hash from front end, need to save as a new row
if (isNewTxHash) {
  if (numberOfConfirmations < 65) {...} 
  else { ... }
} else {
  // Transaction hash already present, need to update row
  if (numberOfConfirmations < 65) {...}
  ...
}
```

apps/server/test-i9n/BridgeApp.i9n.ts
```diff
@@ -109,29 +110,31 @@ describe('Bridge Service Integration Tests', () => {
...
 
     // Step 2: Call addSupportedTokens function and mine the block (block 1004)
     await bridgeUpgradeable
       .connect(defaultAdminSigner)
       .addSupportedTokens(ethers.constants.AddressZero, ethers.utils.parseEther('10'), Math.floor(Date.now() / 1000));
     await hardhatNetwork.generate(1);
 
-    // Step 3: Call bridgeToDeFiChain(_defiAddress, _tokenAddress, _amount) function and mine the block (block 1005)
-    await bridgeUpgradeable.bridgeToDeFiChain(
+    // Step 3: Call bridgeToDeFiChain(_defiAddress, _tokenAddress, _amount) function and mine the block (event emitted at block 1005)
```
weiyuan95, line 136: IMO we don't need to mention the specific block number in the comment here, it doesn't have much meaning

weiyuan95 (changes requested):

apps/server/src/AppService.ts
```diff
@@ -75,6 +74,7 @@ export class AppService {
       },
       data: {
         status: 'CONFIRMED',
+        updatedAt: new Date(),
```
weiyuan95, line 77: I think this is updated automatically as per [this](https://www.prisma.io/docs/reference/api-reference/prisma-schema-reference#updatedat), there's no need to update it manually.

apps/server/src/AppController.ts
```diff
@@ -17,8 +17,8 @@ export class AppController {
     return this.appService.getBalance(address);
   }
 
-  @Get('getAllEventsFromBlockNumber')
-  async getAllEventsFromBlockNumber(@Query('blockNumber') blockNumber: number): Promise<Event[]> {
-    return this.appService.getAllEventsFromBlockNumber(Number(blockNumber));
+  @Post('handleTransaction')
+  async handleTransaction(@Body('transactionHash') transactionHash: string): Promise<boolean> {
```
weiyuan95, line 21: Opened an issue on #334 to ensure that we validate API params.

apps/server/src/EthereumTransactionValidation.pipe.ts
```diff
@@ -0,0 +1,16 @@
+import { BadRequestException, Injectable, PipeTransform } from '@nestjs/common';
+import { ethers } from 'ethers';
+
+@Injectable()
+export class EthereumTransactionValidationPipe implements PipeTransform<string> {
+  async transform(transactionHash: string) {
+    const isTransactionHash =
+      ethers.utils.isHexString(transactionHash) &&
+      transactionHash.length === 66 &&
+      transactionHash.slice(0, 2) === '0x';
```
weiyuan95, line 10: Would rather user the regex here https://ethereum.stackexchange.com/questions/34285/what-is-the-regex-to-validate-an-ethereum-transaction-hash/34286#34286 , it's more succinct IMO

weiyuan95 (changes requested):

apps/server/src/AppConfig.ts
```diff
@@ -1,7 +1,10 @@
 import * as Joi from 'joi';
 
+export const DATABASE_URL = 'DATABASE_URL';
+
 export function appConfig() {
   return {
+    [DATABASE_URL]: process.env[DATABASE_URL],
```
weiyuan95, line 7: This is probably from a bad merge? This looks like it's from an older commit, I think you already made the change to just use `DATABASE_URL: process.env.DATABASE_URL`

weiyuan95 (changes requested):
I think there was a bad merge? Also CI is failing.

pierregee:
Curious if it's a good idea, (good to have!)
1. Return the numberOfConfirmations so the user has an idea in the UI side e.g `15/65 confirmations`
2. Or include this in response ( currentBlockNumber, txReceipt.blockNumber) so the user has an idea in the progress of the tx e.g `8,460,578/8,460,628`

pierregee (changes requested):

weiyuan95:
@pierregee we'll address your comments in another PR, wanna get this through ASAP so that everyone can finally use Prisma 🤣

apps/server/test-i9n/testing/TestingModule.ts
```diff
@@ -16,12 +19,12 @@ export class TestingModule {
...
 
-export function buildTestConfig({
-  startedHardhatContainer,
-  testnet,
-  defichain,
-  dbUrl,
-}: DeepPartial<BuildTestConfigParams> = {}) {
+export function buildTestConfig({ startedHardhatContainer, testnet, defichain, postgres }: BuildTestConfigParams) {
+  if (postgres === undefined) {
+    throw Error('Must pass in started postgres container');
```
weiyuan95, line 24:
```suggestion
    throw Error('Must pass in StartedPostgresContainer');
```

weiyuan95:
Code now is super messy, but will merge this in first to unblock other work.

Issue #277, opened 2023-02-01
chore(contracts): Add `flushFunds` function


weiyuan95:
#### What would you like to be added:
Implement a `flushFunds` function that transfers excess funds across the supported tokens to an address that is hardcoded on the contract. This address should be modifiable, similar to `transactionFee` and `relayerAddress`. Pseudocode below:

```
for each supported token T
  val transferAmount = T.totalBalance - (T.dailyAllowance * 2)
  if transferAmount > 0
    transferTo (hardcodedContract, transferAmount)
```

#### Why is this needed:
Once the funds have reached a particular limit, we want to be able to flush any excess funds to a wallet for operational use.

#### Considerations:
This method does not need to be protected. If someone wants to waste gas to help us flush any excess funds, so be it 😄 The address that the funds are sent to are also hardcoded, so there should be no issues there.

PR #323, opened 2023-02-07
chore(contracts): Add flushFunds function

cuongquangnam:
Implement flushFund function

#277 

Fixes #

#### Developer Checklist:

- [ ] Unit tests\*
- [ ] Added e2e tests\*
- [ ] Added translations\*


packages/smartcontracts/src/tests/BridgeProxy.flushFund.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+    const testToken3 = await ERC20.deploy('Test3', 'T3');
+    await proxyBridge.addSupportedTokens(testToken1.address, toWei('10'), getCurrentTimeStamp());
+    await proxyBridge.addSupportedTokens(testToken2.address, toWei('10'), getCurrentTimeStamp());
+    await proxyBridge.addSupportedTokens(testToken3.address, toWei('10'), getCurrentTimeStamp());
+    await testToken1.mint(proxyBridge.address, toWei('100'));
+    await testToken2.mint(proxyBridge.address, toWei('10'));
+    await testToken3.mint(proxyBridge.address, toWei('100'));
+    await proxyBridge.flushFund();
+    expect(await testToken1.balanceOf(flushReceiveSigner.address)).to.equal(toWei('80'));
+    expect(await testToken2.balanceOf(flushReceiveSigner.address)).to.equal(toWei('0'));
```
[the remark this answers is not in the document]
cuongquangnam: it is the balance of the signer, not the contract, the balance of contract is still 0, haha

packages/smartcontracts/src/contracts/BridgeV1.sol
```diff
@@ -327,6 +358,62 @@ contract BridgeV1 is UUPSUpgradeable, EIP712Upgradeable, AccessControlUpgradeabl
         emit WITHDRAWAL_BY_OWNER(msg.sender, _tokenAddress, amount);
     }
 
+    /**
+     * @notice Function to flush the excess funds across supported tokens to a hardcoded address
+     * anyone can call this function
+     */
+    function flushFund() external {
+        for (uint256 i = 0; i < supportedTokens.length(); i++) {
```
Abhishekkochar, line 366:
```suggestion
        for (uint256 i = 0; i < supportedTokens.length(); ++i) {
```
Will do the same thing, however saves us 5 units. More [here](https://cygaar.substack.com/p/why-i-is-cheaper-than-i-in-solidity)

packages/smartcontracts/src/tests/BridgeProxy.flushFund.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+    expect(await testToken3.balanceOf(flushReceiveSigner.address)).to.equal(toWei('80'));
+  });
+
+  it('Should be able to change flushReceiveAddress', async () => {
+    const { proxyBridge, flushReceiveSigner } = await loadFixture(deployContracts);
+    const ERC20 = await ethers.getContractFactory('TestToken');
```
Abhishekkochar, line 29: We can import `testToken` from loadFixture. We can avoid repetition.

packages/smartcontracts/src/tests/BridgeProxy.flushFund.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+    await proxyBridge.changeFlushReceiveAddress(newFlushReceiveAddress);
+    await testToken1.mint(proxyBridge.address, toWei('100'));
+    await proxyBridge.flushFund();
+    expect(await testToken1.balanceOf(newFlushReceiveAddress)).to.equal(toWei('100'));
+    expect(await testToken1.balanceOf(flushReceiveSigner.address)).to.equal(toWei('80'));
+  });
+
+  it('Should be able to change acceptableRemainingDays', async () => {
+    const { proxyBridge, flushReceiveSigner } = await loadFixture(deployContracts);
+    const ERC20 = await ethers.getContractFactory('TestToken');
```
Abhishekkochar, line 45: Same as mentioned earlier

packages/smartcontracts/src/tests/BridgeProxy.flushFund.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+    await testToken1.mint(proxyBridge.address, toWei('100'));
+    await testToken2.mint(proxyBridge.address, toWei('10'));
+    await testToken3.mint(proxyBridge.address, toWei('100'));
+    await proxyBridge.flushFund();
+    expect(await testToken1.balanceOf(flushReceiveSigner.address)).to.equal(toWei('80'));
+    expect(await testToken2.balanceOf(flushReceiveSigner.address)).to.equal(toWei('0'));
+    expect(await testToken3.balanceOf(flushReceiveSigner.address)).to.equal(toWei('80'));
+  });
+
+  it('Should be able to change flushReceiveAddress', async () => {
```
Abhishekkochar, line 27: It will be good, if we can get similar to [this](https://github.com/WavesHQ/bridge/blob/main/packages/smartcontracts/src/tests/BridgeProxy.Withdraw.spec.ts#L8). Sanity check only admins can call this function.

packages/smartcontracts/src/tests/BridgeProxy.flushFund.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+    const { proxyBridge, flushReceiveSigner } = await loadFixture(deployContracts);
+    const ERC20 = await ethers.getContractFactory('TestToken');
+    const testToken1 = await ERC20.deploy('Test1', 'T1');
+    await proxyBridge.addSupportedTokens(testToken1.address, toWei('10'), getCurrentTimeStamp());
+    await testToken1.mint(proxyBridge.address, toWei('100'));
+    await proxyBridge.flushFund();
+    expect(await testToken1.balanceOf(flushReceiveSigner.address)).to.equal(toWei('80'));
+    const newFlushReceiveAddress = (await ethers.provider.listAccounts())[4];
+    await proxyBridge.changeFlushReceiveAddress(newFlushReceiveAddress);
```
Abhishekkochar, line 36:
```suggestion
    await proxyBridge.connect(defaultAdmin).changeFlushReceiveAddress(newFlushReceiveAddress);
```

Abhishekkochar (changes requested):

apps/server/package.json
```diff
@@ -52,6 +52,7 @@
     "ethers": "^5.7.2",
     "joi": "^17.7.0",
     "nestjs-pino": "^3.1.2",
+    "next": "^13.1.6",
```
weiyuan95, line 55: Not intended?🤔

apps/web/package.json
```diff
@@ -27,6 +27,7 @@
     "connectkit": "^1.1.3",
     "dayjs": "^1.11.7",
     "ethers": "^5.7.2",
+    "next": "^13.1.6",
```
weiyuan95, line 30: Same here

packages/smartcontracts/src/contracts/BridgeV1.sol
```diff
@@ -326,6 +357,65 @@ contract BridgeV1 is UUPSUpgradeable, EIP712Upgradeable, AccessControlUpgradeabl
...
+            address supToken = supportedTokens.at(i);
+            if (
+                (IERC20(supToken).balanceOf(address(this)) >
+                    acceptableRemainingDays * tokenAllowances[supToken].dailyAllowance) &&
+                // the same analysis as in bridgeToDeFiChain applies here, if we are still
+                // in the period for addingSupport or changingDailyAllowane, we will not flush for the token
```
weiyuan95, line 371:
```suggestion
                // the same logic as in bridgeToDeFiChain applies here, if we are still
                // in the period for addingSupport or changingDailyAllowance, we will not flush for that particular token
```

packages/smartcontracts/src/contracts/BridgeV1.sol
```diff
@@ -326,6 +357,65 @@ contract BridgeV1 is UUPSUpgradeable, EIP712Upgradeable, AccessControlUpgradeabl
...
+     */
+    function changeAcceptableRemainingDays(uint256 _newAcceptableRemainingDays) external {
+        if (!checkRoles()) revert NON_AUTHORIZED_ADDRESS();
+        uint256 _oldAcceptableRemainingDays = acceptableRemainingDays;
+        acceptableRemainingDays = _newAcceptableRemainingDays;
+        emit CHANGE_ACCEPTABLE_REMAINING_DAYS(_oldAcceptableRemainingDays, _newAcceptableRemainingDays);
+    }
+
+    /**
+     * @notice to get the supported tokens, as recursive data structure (supportedTokens) cannot be made public
```
weiyuan95, line 406:
```suggestion
     * @notice to get the ERC20 addresses of supported tokens, as recursive data structure (supportedTokens) cannot be made public
```
Some questions:
- What does it mean by recursive data structure? Isn't this just the values of the set?
- Why cant `supportTokens` be public? 🤔 Some limitation of the utility?

packages/smartcontracts/src/tests/BridgeProxy.flushFund.spec.ts
```diff
@@ -0,0 +1,143 @@
...
+    expect(balanceAfterFlush).to.equal(balanceBeforeFlush);
+  });
+
+  it('Should not flush fund for a token if it is in the change allowance period', async () => {
+    const { proxyBridge, flushReceiveSigner, testToken, defaultAdminSigner } = await loadFixture(deployContracts);
+    // latestResetTimestamp = block.timestamp + 59 = X + 59
+    await proxyBridge.addSupportedTokens(testToken.address, toWei('10'), (await time.latest()) + 60);
+    const latestResetTimestampAfterAddingSupport = (await proxyBridge.tokenAllowances(testToken.address))
+      .latestResetTimestamp;
+    expect(latestResetTimestampAfterAddingSupport).to.equal((await time.latest()) + 59);
```
weiyuan95, line 109:
```suggestion
    const expectedTime = getCurrentTimeStamp({ additionalTime: 60 });
    // latestResetTimestamp = block.timestamp + 59 = X + 59
    await proxyBridge.addSupportedTokens(testToken.address, toWei('10'), expectedTime);
    const latestResetTimestampAfterAddingSupport = (await proxyBridge.tokenAllowances(testToken.address))
      .latestResetTimestamp;
    expect(latestResetTimestampAfterAddingSupport).to.equal(expectedTime);
```

I'm guessing you were previously calling `getCurrentTimeStamp` multiple times, which was causing some time errors. If we did it in this manner with a constant reference point for the time, I don't think we'll have the problem of wrong timestamps (hardhat's `time.now()` vs node's `Date.now()`)

We can adjust the tests at the later date though - the logic still looks sound to me.

packages/smartcontracts/src/tests/BridgeProxy.flushFund.spec.ts
```diff
@@ -0,0 +1,143 @@
...
+    );
+    const balanceBefore1stFlush = await testToken.balanceOf(flushReceiveSigner.address);
+    // block.timestamp = Y + 1 < Y + 1 days = new latestResetTimestamp
+    await proxyBridge.flushFund();
+    const balanceAfter1stFlush = await testToken.balanceOf(flushReceiveSigner.address);
+    expect(balanceAfter1stFlush).to.equal(balanceBefore1stFlush);
+    // new block with block.timestamp = Y + 1 + ONE_DAY
+    await time.increase(ONE_DAY);
+    const balanceBefore2ndFlush = await testToken.balanceOf(flushReceiveSigner.address);
+    const balanceBridgeBefore2ndFlush = await testToken.balanceOf(proxyBridge.address);
```
weiyuan95, line 133:
```suggestion
    const bridgeBalanceBefore2ndFlush = await testToken.balanceOf(proxyBridge.address);
```

Same for the other variables as well.

packages/smartcontracts/src/tests/BridgeProxy.flushFund.spec.ts
```diff
@@ -0,0 +1,143 @@
...
+    // block.timestamp = X + 1
+    await testToken.mint(proxyBridge.address, toWei('100'));
+    const balanceBeforeFlush = await testToken.balanceOf(flushReceiveSigner.address);
+    // block.timestamp = X + 2
+    await proxyBridge.flushFund();
+    const balanceAfterFlush = await testToken.balanceOf(flushReceiveSigner.address);
+    expect(balanceAfterFlush).to.equal(balanceBeforeFlush);
+  });
+
+  it('Should not flush fund for a token if it is in the change allowance period', async () => {
```
weiyuan95, line 103: Since this test is a lot more complex, it'll be good to have some comments for it. For example
```ts
// mint 100 tokens to bridge
...
...
// Change dailyAllowance of bridge
// Call flush funds
// Funds should not be flushed since it is not officially supported yet
// Advance time till token is supported
// Flush funds
// Bridge should have correct balance after correct amount is flushed (20 * 2)
```
Would help to break up the code and expectations into blocks that are easier to read and helps the user to understand the test too.

It'll also be good to have some math to help readers keep track, for eg
```ts
// 40 (current amount on bridge) + 100 (newly minted) - 20 (flushed amount)
```

weiyuan95 (changes requested):
IMO the logic is sound, we can work on bolstering the test cases later.

apps/web/package.json
cuongquangnam: Oh, I was trying to add this when the hooks before my commit broke. It said next is not found ...

packages/smartcontracts/src/contracts/BridgeV1.sol
cuongquangnam: The compiler will fail when you make supportedTokens public: 
 `TypeError: Internal or recursive type is not allowed for public state variables.`
My guess at that point is that supportedTokens, which is of type AddressSet ( https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/structs/EnumerableSet.sol#L233) and this type is recursive ( a struct inside a struct)

packages/smartcontracts/src/tests/BridgeProxy.flushFund.spec.ts
weiyuan95: Problem has been figured out. Opened https://github.com/WavesHQ/bridge/issues/357 to make sure that we use the fix approach as well for the other tests instead of the current `getCurrentTimeStamp` as well.

packages/smartcontracts/src/tests/BridgeProxy.flushFund.spec.ts
weiyuan95: With the current approach, remember to make a note on how much the timestamp advances as well 😛

apps/server/package.json
cuongquangnam: Reverted changes on this

apps/web/package.json
cuongquangnam: Reverted changes on this

weiyuan95 (approved):
Approving and merging this in. The Hardhat test is a bit messy right now, but it will be tackled in #357 . It should not have any impact on the actual contract logic.

PR #324, opened 2023-02-07
feat(server): sign claims data

lykalabrada:
Added API to signs claims and return signature and nonce. This API will called before user can `claimFund`.
Another PR for frontend to integrate this API will be raised later.

Fixes #

#### Developer Checklist:

- [ ] Unit tests\*
- [ ] Added e2e tests\*
- [ ] Added translations\*


apps/server/src/AppService.ts
```diff
@@ -1,16 +1,22 @@
-import { Inject, Injectable } from '@nestjs/common';
+import { HttpException, HttpStatus, Inject, Injectable } from '@nestjs/common';
 import { ConfigService } from '@nestjs/config';
 import { BigNumber, Contract, ethers, Event } from 'ethers';
 import { BridgeV1__factory } from 'smartcontracts';
 
 import { ETHERS_RPC_PROVIDER } from './modules/EthersModule';
 
+interface SignClaim {
```
weiyuan95, line 8: Minor nit here but I would rather put this on the bottom of the file 😛

apps/server/src/AppService.ts
```diff
@@ -1,16 +1,22 @@
...
+  amount: string;
+}
+
 @Injectable()
 export class AppService {
   private contract: Contract;
 
   constructor(
-    @Inject(ETHERS_RPC_PROVIDER) readonly ethersRpcProvider: ethers.providers.StaticJsonRpcProvider,
+    @Inject(ETHERS_RPC_PROVIDER) readonly ethersRpcProvider: ethers.providers.JsonRpcProvider,
```
weiyuan95, line 19: Hmm, I would understand if this is necessary on the front end (since the network/url selected by the user is not static on metamask) but why is it non static on the server? Wouldn't the RPC url (and hence network) always be the same?

apps/server/src/AppService.ts
```diff
@@ -33,4 +39,54 @@ export class AppService {
     const eventSignature = this.contract.filters.BRIDGE_TO_DEFI_CHAIN();
     return this.contract.queryFilter(eventSignature, blockNumber, currentBlockNumber - 65);
   }
+
+  async signClaim({ receiverAddress, tokenAddress, amount }: SignClaim): Promise<{ signature: string; nonce: number }> {
+    try {
+      // Connect signer ETH wallet (admin/operational wallet)
+      const wallet = new ethers.Wallet(this.configService.getOrThrow('ethereum.testnet.ethWalletPrivKey'));
+      const signer = wallet.connect(this.ethersRpcProvider);
```
weiyuan95, line 47:
```suggestion
      const wallet = new ethers.Wallet(this.configService.getOrThrow('ethereum.testnet.ethWalletPrivKey'), this.ethersRpcProvider);
```

We can skip the redundant `connect` call by passing it in directly to the constructor.

apps/server/src/AppService.ts
```diff
@@ -33,4 +39,54 @@ export class AppService {
...
+          { name: 'nonce', type: 'uint256' },
+          { name: 'deadline', type: 'uint256' },
+          { name: 'tokenAddress', type: 'address' },
+        ],
+      };
+      const data = {
+        to: receiverAddress,
+        amount,
+        nonce,
+        deadline: ethers.constants.MaxUint256,
```
weiyuan95, line 72: The deadline should be 2359 GMT of the current day, but i'm guessing that will be an implementation that might come later 😛

apps/web/src/config/ABIs/testnet/BridgeV1.json
```diff
@@ -0,0 +1,641 @@
+[
+  { "inputs": [], "name": "DO_NOT_SEND_ETHER_WITH_ERC20", "type": "error" },
```
weiyuan95, line 2: Instead of hardcoding this,`app/web` can have a dependency on `"smartcontracts": "workspace:0.0.0",` so that we can access the ABI with `BridgeV1__factory.abi`. I will also always be the correct version, no need to manually update!

weiyuan95 (changes requested):
Left some comments on changes related to the server 😄

pierregee:
Sorry @lykalabrada! I think it was automerged when I was debugging it on `verify-funds` branch and pushed it! Thanks for reopening and pointing to main

apps/server/src/AppService.ts
lykalabrada: Sry I was just testing around and forgot to revert this, thanks for checking!

apps/web/src/config/ABIs/testnet/BridgeV1.json
lykalabrada: Good point! Since this is on the web side we will update this on a completely separate PR.
For the other comments above, fixed on [this PR](https://github.com/WavesHQ/bridge/pull/344) 🙏

PR #330, opened 2023-02-08
feat(core): verify funds - DFC -> EVM

pierregee:
- [x] Create /verify endpoint to verify if the balance is in the wallet's address
- [x] Handle edge cases - display custom error code in UI to help debugging 
- [x] Add e2e tests

Fixes #

#### Additional comments?:
These are tasks that will be handled in another PR
- Check if daily limit has not been reached - blocked by the task to retrieve how much is already used 
- Call signClaims inside verify (will handle in another PR), also dependent to https://github.com/WavesHQ/bridge/pull/344
- Insert record of the transaction in the db to be able to track the daily limit

#### Developer Checklist:

- [ ] Unit tests\*
- [x] Added e2e tests\*
- [ ] Added translations\*


apps/server/src/defichain/model/VerifyDto.ts
```diff
@@ -0,0 +1,18 @@
+import { IsString } from 'class-validator';
+
+export class VerifyDto {
+  @IsString()
```
weiyuan95, line 4: Can probably use `IsNumberString`, more decorators [here](https://github.com/typestack/class-validator#validation-decorators). Then we can also check for things like is the number positive, etc.

apps/server/src/defichain/model/VerifyDto.ts
```diff
@@ -0,0 +1,18 @@
...
+
+  @IsString()
+  symbol: string;
+
+  constructor(address: string, amount: string, symbol: string) {
+    this.address = address;
+    this.amount = amount;
+    this.symbol = symbol;
+  }
+}
```
weiyuan95, line 18: Since this class only exists for data validation and transfer, we can have a data class which represents the object we want to deal with in our server. For eg.
```ts
class VerifyObject {
  constructor(public readonly amount: BigNumber, ...) { }
}
```
and then we have 

```ts
class VerifyDto {
  // ...
  // same as current implementation
  // ...

  toObj(): VerifyObject {
    return new VerifyObject(new BigNumber(this.amount), ...)
  }
}
```

And then we work with `VerifyObject` in our code instead of `VerifyDto`

apps/server/test-i9n/defichain/verify.i9n.ts
```diff
@@ -0,0 +1,231 @@
...
+    localAddress = await wallet.getAddress();
+  });
+
+  afterAll(async () => {
+    // teardown database
+    await startedPostgresContainer.stop();
+    await testing.stop();
+  });
+
+  it('should throw error if symbol is not valid', async () => {
```
weiyuan95, line 62: We can actually write more succinct validation tests with `it.each` (basically a for loop for jest's `it`). For example in this case:

```ts
// needs some setup
type MockedPayload = {
  amount: string,
  symbol: string,
  address: string
}
function mockPayload(params: Partial<MockedPayload>): MockedPayload {
  return {
    amount: '1',
    symbol: 'BTC',
    address: 'abc1234',
    ...params // overrides any of the 'right' values
  }
}

it.each([
  mockPayload({address: '10923'})
  // ...
])(`Should fail request with invalid request %p`, async(mockedPayload) => {
    // use mock payload
    // in this loop mockedPayload == {amount: '1', symbol: 'btc', address: '10923'}
})

```

weiyuan95 (changes requested):

apps/server/test-i9n/defichain/verify.i9n.ts
pierregee: Grouping it in `.each` is tempting! But looping it in such way won't be able to run some code prerequisites unique to each test like [1](https://github.com/WavesHQ/bridge/pull/330/files#diff-48b02ff748a2982153d8aefeeaddb3befc212ca03ba354f843b8bbb864b6f23fR112-R125) [2](https://github.com/WavesHQ/bridge/pull/330/files#diff-48b02ff748a2982153d8aefeeaddb3befc212ca03ba354f843b8bbb864b6f23fR131-R150) 🤔

pierregee: Tried it still and it seems the first param of `it.each()` isn't waiting for "beforeAll" so some vars can't be retrieved

pierregee: To make it "at least" more succinct, how about putting the api call in a function? Helps a bit..
``` 
async function verify(mockedPayload: MockedPayload) {
    const initialResponse = await testing.inject({
      method: 'POST',
      url: `${VERIFY_ENDPOINT}`,
      payload: mockedPayload,
    });
    const response = JSON.parse(initialResponse.body);

    return response;
  }


it('should throw error if symbol is not valid', async () => {
  const initialResponse = await verify({
    amount: '1',
    symbol: '_invalid_symbol_',
    address: localAddress,
  });
  expect(initialResponse).toStrictEqual({ isValid: false, statusCode: CustomErrorCodes.TokenSymbolNotValid });
});
```

weiyuan95: > Tried it still and it seems the first param of it.each() isn't waiting for "beforeAll" so some vars can't be retrieved

Ah ok, we need the `localAddress` to actually confirm that the verify flow works, and it cannot be any generic valid address since i'm guessing that won't be in the DB.

Yep, I think what you've suggested works as well, since we don't really care about the `inject` and `response` code, what's important is the `params` used to call the endpoint 😬

apps/server/src/defichain/controllers/WhaleWalletController.ts
```diff
@@ -1,21 +1,29 @@
...
 @Controller('/wallet')
 export class WhaleWalletController {
   private network: EnvironmentNetwork;
 
   constructor(private readonly whaleWalletService: WhaleWalletService, private readonly configService: ConfigService) {
-    this.network = configService.getOrThrow<EnvironmentNetwork>(`defichain.network`);
+    this.network = this.configService.getOrThrow<EnvironmentNetwork>(`defichain.network`);
   }
 
-  @Throttle(5, 60)
```
thedoublejay, line 16: Why we removed this?

apps/server/src/defichain/services/WhaleWalletService.ts
```diff
@@ -1,12 +1,87 @@
...
+import { VerifyDto } from '../model/VerifyDto';
 import { WhaleWalletProvider } from '../providers/WhaleWalletProvider';
 
 @Injectable()
 export class WhaleWalletService {
   constructor(private readonly whaleWalletProvider: WhaleWalletProvider, private prisma: PrismaService) {}
 
+  async verify(
+    verify: VerifyDto,
+    network: EnvironmentNetwork = EnvironmentNetwork.MainNet,
```
thedoublejay, line 18: This shouldn't be optional

thedoublejay: Since we're already requiring it on the client, I think just remove the default values

thedoublejay (changes requested):

apps/server/src/defichain/controllers/WhaleWalletController.ts
weiyuan95: Starting to think that the throttle should just be set globally instead of on a controller level 🤔

thedoublejay: But some endpoints should have stricter throttles, like this which stores new data on the DB

thedoublejay: While others can be more lenient

apps/server/src/defichain/model/VerifyDto.ts
pierregee: I'm not quite sure how this would work in the controller level 🤔 
```
  @Post('verify')
  async verify(@Body() body: VerifyDto): Promise<{ isValid: boolean; statusCode?: CustomErrorCodes }> {

    // doesn't work
    return this.whaleWalletService.verify(
      body.toObj(), // won't work because toObj is undefined
      this.network,
    );
  
    // this works
    return this.whaleWalletService.verify(
      new VerifyObject(new BigNumber(body.amount), body.address, body.symbol),
      this.network,
    );
  }
```

weiyuan95: Ah sorry, mixing up implementations across different repos. We'll need to `transform` the object since it's just a plan JS object over the wire.

https://docs.nestjs.com/techniques/validation#transform-payload-objects

apps/server/src/defichain/controllers/WhaleWalletController.ts
pierregee: Ah ok sure, moved it back here to endpoint-level! I have to use a [config](https://github.com/WavesHQ/bridge/pull/330/files#diff-9bb3e234e20f4e46dad3d85462312a05f11e9c3b201e4233009c2a92b98bc153R33) to be able to test it. I haven't deep dive yet if there's a NestJS way of configuring the limit in e2e tests so I used config + jest.mock

apps/server/test-i9n/defichain/verify.i9n.ts
pierregee: Yes, implemented the verify function above instead! Lemmeknow if it can be improved 🙏

apps/server/src/defichain/model/VerifyDto.ts
pierregee: This works! Thanks for sharing

Issue #319, opened 2023-02-07
chore(config): Config files should be scoped to one network


weiyuan95:
#### What would you like to be added:
As per title. There's no need for the app to have configs for the different networks, when they can be scoped to a single network (or network pair).

For example
- 1 Server will be deployed with the ethereum network pointing the goerli and the defichain network pointing to playground
- Another server will be deployed with the ethereum network point to mainnet and defichain network pointing to mainnet

#### Why is this needed:
- Prevent single point of failure
- Prevent a 'super' app that messes up devops processes

weiyuan95:
#331 handles the consolidation on the DFC side, we'll still need to do it for the Ethereum side.

PR #331, opened 2023-02-08
feature(apps/server): DeFiChain - set network to config

thedoublejay:
Allows for using a single network on the server. Will not support multiple network switching to make deployed server consistent with a single network.

Partial fix for https://github.com/WavesHQ/bridge/issues/319

#### Developer Checklist:

- [ ] Unit tests\*
- [ ] Added e2e tests\*
- [ ] Added translations\*


apps/server/src/AppConfig.ts
```diff
@@ -4,10 +4,9 @@ import * as Joi from 'joi';
 export function appConfig() {
   return {
     defichain: {
-      [EnvironmentNetwork.MainNet]: process.env.DEFICHAIN_MAINNET_KEY,
-      [EnvironmentNetwork.RemotePlayground]: process.env.DEFICHAIN_REMOTE_KEY,
-      [EnvironmentNetwork.LocalPlayground]: process.env.DEFICHAIN_LOCAL_KEY,
-      localWhaleURL: 'http://localhost:19553',
+      key: process.env.DEFICHAIN_MAINNET_KEY,
+      whaleURL: process.env.DEFICHAIN_WHALE_URL || 'http://localhost:19553',
```
weiyuan95, line 8: Since we are already using a `.env` file, we might want to keep hardcoded defaults there.

weiyuan95: Also, if we want a default value for something that is `undefined`, we should use the [null coalescing operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing) instead (`??`). `||` checks for `falsy` values.

apps/server/src/defichain/NetworkContainer.ts
```diff
@@ -0,0 +1,10 @@
+import { ConfigService } from '@nestjs/config';
+import { EnvironmentNetwork } from '@waveshq/walletkit-core';
+
+export class NetworkContainer {
+  protected network: EnvironmentNetwork;
+
+  constructor(configService: ConfigService) {
+    this.network = configService.getOrThrow<EnvironmentNetwork>(`defichain.network`);
```
weiyuan95, line 8: If we only need this `network` value, we should be getting it from `configService` (dependency injection) instead of using inheritance like this. Another way is to `provide` and inject the value similar to [this](https://docs.nestjs.com/fundamentals/custom-providers#non-class-based-provider-tokens).

I'm not sure if there will be other network related logic though, if there are we can consider leaving in this abstraction.

apps/server/src/defichain/services/DeFiChainTransactionService.ts
```diff
@@ -13,50 +15,48 @@ const MAX_TIMEOUT = 300000;
...
-export class DeFiChainTransactionService {
+export class DeFiChainTransactionService extends NetworkContainer {
   constructor(
     private readonly whaleWalletProvider: WhaleWalletProvider,
     private readonly clientProvider: WhaleApiClientProvider,
     private readonly whaleClient: WhaleApiService,
-  ) {}
+    private readonly configService: ConfigService,
+  ) {
+    super(configService);
```
weiyuan95, line 25: Yep the more I see it, the more it looks like an unnecessary abstraction. We can just declare `this.network` here instead for use in the class.

weiyuan95 (changes requested):
Quick comments

apps/server/src/defichain/NetworkContainer.ts
thedoublejay: @weiyuan95 Updated! I wanted to use it directly but was thinking to repeat codes on each class. Anyway if it's only a network property, we can just use DI from `configService`.

weiyuan95: Yep, IMO i would rather have some repeating code than an unnecessary abstraction 😬

apps/server/.env
```diff
@@ -4,3 +4,6 @@ POSTGRES_DB="bridge"
 POSTGRES_USER="playground"
 POSTGRES_PASSWORD="playground"
 DATABASE_URL="postgresql://${POSTGRES_USER}:${POSTGRES_PASSWORD}@localhost:5432/${POSTGRES_DB}?schema=public"
+ETHEREUM_RPC_URL="localhost:8545"
```
weiyuan95, line 7:
```suggestion
ETHEREUM_RPC_URL="http://localhost:8545"
```

weiyuan95: Also we might need to use `http://127.0.0.1:8545` to pass the JOI validation

weiyuan95 (approved):

PR #346, opened 2023-02-09
feat(Contracts): Added name and version constants to contract

### LINKED ISSUE #345: `Contracts`: Adding name and version in contract as constants
With the current implementation, there are no mentioned of name and version of the contract. 
For the backend, devs have to manually change the `env` var if upgrading the contract later.

Abhishekkochar:
As per the title, adding the constants to SM.
Easy for backend to fetch this data straight from contract instead of manually updating.
Fixes #345 

#### Developer Checklist:

- [ ] Unit tests\*
- [ ] Added e2e tests\*
- [ ] Added translations\*

### DIFF
--- a/packages/smartcontracts/src/contracts/BridgeV1.sol
+++ b/packages/smartcontracts/src/contracts/BridgeV1.sol
@@ -91,6 +91,9 @@ contract BridgeV1 is UUPSUpgradeable, EIP712Upgradeable, AccessControlUpgradeabl
 
     bytes32 public constant OPERATIONAL_ROLE = keccak256('OPERATIONAL_ROLE');
 
+    string public constant name = 'QUANTUM_BRIDGE';
+    string public constant version = '1.0';
+
     // Initial Tx fee 0.3%. Based on dps (e.g 1% == 100dps)
     uint256 public transactionFee;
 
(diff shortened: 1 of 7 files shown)


weiyuan95 (approved):

packages/smartcontracts/src/contracts/BridgeV1.sol
```diff
@@ -91,6 +91,9 @@ contract BridgeV1 is UUPSUpgradeable, EIP712Upgradeable, AccessControlUpgradeabl
 
     bytes32 public constant OPERATIONAL_ROLE = keccak256('OPERATIONAL_ROLE');
 
+    string public constant name = 'QUANTUM_BRIDGE';
+    string public constant version = '1.0';
```
weiyuan95, line 95: We need to pass this to `initialize` instead of using the current method were these values are passed in as parameters

weiyuan95 (changes requested):

packages/smartcontracts/src/contracts/BridgeV1.sol
```diff
@@ -173,23 +176,19 @@ contract BridgeV1 is UUPSUpgradeable, EIP712Upgradeable, AccessControlUpgradeabl
...
-        string calldata _name,
-        string calldata _version,
         address _initialAdmin,
         address _initialOperational,
         address _relayerAddress,
         uint256 _fee
     ) external initializer {
         __UUPSUpgradeable_init();
-        __EIP712_init(_name, _version);
+        __EIP712_init(name, version);
```
Abhishekkochar, line 191: @weiyuan95, we can pass it as constant values.

weiyuan95 (approved):

PR #367, opened 2023-02-13
chore(server): Separation of EVM Transaction Confirmer logic from generic App* classes

### LINKED ISSUE #347: chore(server): Code cleanup
# This is not important for launch, mostly a code cleanup task

As per title. These tasks should be in separate PRs.
- [x] Remove the unnecessary and generic `App*` classes and replace them classes that have their own areas of concern.
- [x] Test cleanup
- [x] Address other comments in #310

#### Why is this needed:
Currently, there, the logic related to the confirming EVM transaction sits on `AppController` and `AppService`. We shouldn't need such generic controllers and services, they should be properly scoped into their own areas of concern. For eg. `EvmTransactionController`.

greggtan:
Separated Ethereum Transaction Confirmer service from app module, controller and service logic.

Fixes #347

#### Developer Checklist:


apps/server/src/ethereum/controllers/EthereumController.ts
```diff
@@ -0,0 +1,16 @@
+import { Body, Controller, Post } from '@nestjs/common';
+
+import { EthereumTransactionValidationPipe } from '../../pipes/EthereumTransactionValidation.pipe';
+import { EVMTransactionConfirmerService } from '../services/EVMTransactionConfirmerService';
+
+@Controller('')
```
weiyuan95, line 6:
```suggestion
@Controller()
```

apps/server/src/AppService.ts
```diff
@@ -1,26 +1,11 @@
...
-    private configService: ConfigService,
-    private prisma: PrismaService,
-  ) {
-    this.contract = new ethers.Contract(
-      this.configService.getOrThrow('ethereum.contracts.bridgeProxy.address'),
-      BridgeV1__factory.abi,
-      this.ethersRpcProvider,
-    );
-  }
+  constructor(@Inject(ETHERS_RPC_PROVIDER) readonly ethersRpcProvider: ethers.providers.StaticJsonRpcProvider) {}
```
weiyuan95, line 8: If this class is not being used anywhere (I don't think it is?) let's just delete it

apps/server/src/AppController.ts
```diff
@@ -17,11 +16,4 @@ export class AppController {
   async getBalance(@Query('address') address: string): Promise<BigNumber> {
```
weiyuan95, line 16: Same here, if it's not being used let's delete. It's ok to have an `AppModule` (root level module for the app that consolidates all modules that the app needs) but if we don't need a controller/service then we can remove the unnecessary parts

weiyuan95 (changes requested):
Some quick comments

weiyuan95:
LGTM, just need to resolve the conflicts

PR #372, opened 2023-02-13
chore(contracts): Implement TestBridge for QA testing

### LINKED ISSUE #366: chore(contracts): Implement `TestBridgeV1` contract
Implement a contract for testing purposs, where instead of a 24 hour reset time, the contract state (limit refresh, etc) will refresh every 1 hour instead. We don't need any robust testing for this, but some sanity checks will be good.

Once this gets implemented, we should upgrade the current testnet contract to this version instead.

#### Why is this needed:
To make it easier for the QA team to test against testnet.

Abhishekkochar:
For testnet only. With the current contract, reset time is set to a day. While this work great, however, for testing purpose we  set to an hour instead.  
#366 

Fixes #366

#### Developer Checklist:

- [ ] Unit tests\*
- [ ] Added e2e tests\*
- [ ] Added translations\*


packages/smartcontracts/src/contracts/BridgeV2.sol
```diff
@@ -0,0 +1,483 @@
...
+/* This error occurs when `_newResetTimeStamp` is before block.timestamp
+*/
+error EXPIRED_CLAIM();
+
+/** @notice @dev 
+/* This error occurs when `_amount` is zero
+*/
+error AMOUNT_CAN_NOT_BE_ZERO();
+
+contract BridgeV2 is UUPSUpgradeable, EIP712Upgradeable, AccessControlUpgradeable {
```
weiyuan95, line 72: Thinking if we call it `TestBridge` instead so that we don't accidentally mess up 😛

packages/smartcontracts/src/contracts/BridgeV2.sol
```diff
@@ -0,0 +1,483 @@
...
+        bytes calldata _defiAddress,
+        address _tokenAddress,
+        uint256 _amount
+    ) external {
+        if (!supportedTokens.contains(_tokenAddress)) revert TOKEN_NOT_SUPPORTED();
+        uint256 tokenAllowanceStartTime = tokenAllowances[_tokenAddress].latestResetTimestamp;
+        if (block.timestamp < tokenAllowanceStartTime) revert STILL_IN_CHANGE_ALLOWANCE_PERIOD();
+        if (_amount == 0) revert AMOUNT_CAN_NOT_BE_ZERO();
+        // Transaction is within the last tracked day's daily allowance
+        if (tokenAllowances[_tokenAddress].latestResetTimestamp + (3600) > block.timestamp) {
```
weiyuan95, line 286: Oh hmm, we can't use `1 hours`? 🤔

packages/smartcontracts/src/contracts/BridgeV2.sol
```diff
@@ -0,0 +1,483 @@
...
+    function bridgeToDeFiChain(
+        bytes calldata _defiAddress,
+        address _tokenAddress,
+        uint256 _amount
+    ) external {
+        if (!supportedTokens.contains(_tokenAddress)) revert TOKEN_NOT_SUPPORTED();
+        uint256 tokenAllowanceStartTime = tokenAllowances[_tokenAddress].latestResetTimestamp;
+        if (block.timestamp < tokenAllowanceStartTime) revert STILL_IN_CHANGE_ALLOWANCE_PERIOD();
+        if (_amount == 0) revert AMOUNT_CAN_NOT_BE_ZERO();
+        // Transaction is within the last tracked day's daily allowance
```
weiyuan95, line 285:
```suggestion
        // Transaction is within the last tracked hour's daily allowance
```

packages/smartcontracts/src/contracts/BridgeV2.sol
Abhishekkochar: Oh I checked, we can.

weiyuan95 (approved):
LGTM, @cuongquangnam do you want to do a sanity check?

PR #376, opened 2023-02-13
chore(server): Refactor for EthereumTransactionConfirmerService

### LINKED ISSUE #347: chore(server): Code cleanup

greggtan:
Separated Ethereum Transaction Confirmer service from app module, controller and service logic.

Fixes #347

#### Developer Checklist:


weiyuan95 (approved):

weiyuan95:
Changes brought up by Pierre in the previous PR will be handled in another PR

PR #387, opened 2023-02-14
feat(EVMTransactionConfirmer): Return number of confirmations for UI

greggtan:

Changes the output for /ethereum/handleTransaction route to return number of confirmations and whether the transaction is confirmed.

Fixes #

#### Developer Checklist:

- [x] Unit tests\*
- [x] Added e2e tests\*

### DIFF
--- a/apps/server/src/types/HandledEVMTransaction.ts
+++ b/apps/server/src/types/HandledEVMTransaction.ts
@@ -1,0 +1,4 @@
+export interface HandledEVMTransaction {
+  numberOfConfirmations: number;
+  isConfirmed: boolean;
+}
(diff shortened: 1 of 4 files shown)


apps/server/src/types/HandledEVMTransaction.ts
```diff
@@ -0,0 +1,4 @@
+export interface HandledEVMTransaction {
```
weiyuan95, line 1: This can sit in `EvmTransactionConfirmerService`. I'm also not too sure if it even needs to be exported, I don't think any class will import this 🤔

weiyuan95:
@pierregee this addresses your comment in https://github.com/WavesHQ/bridge/pull/310#issuecomment-1425130435

weiyuan95 (approved):
Will leave the final review for Pierre 😬

lykalabrada (approved):

PR #397, opened 2023-02-15
chore(Contracts): Testing proxy behaviour

Abhishekkochar:
Testing with two different modifier `reinitializer()` and `initializer`

If we need to init an upgraded new contract, such action will not be doable if the initialize function have the `initializer` modifier. We will need `reinitializer()` in this case.

Fixes #

#### Developer Checklist:

- [ ] Unit tests\*
- [ ] Added e2e tests\*
- [ ] Added translations\*


packages/smartcontracts/src/tests/BridgeProxy.proxyBehaviour.spec.ts
```diff
@@ -0,0 +1,160 @@
...
+      operationalAdminSigner.address,
+      // community wallet address
+      communityAddress,
+      30,
+      communityAddress,
+      2,
+    ]);
+
+    // upgrade to the new contract
+    await expect(proxyBridge.upgradeToAndCall(TestContract.address, encodedData)).to.be.rejectedWith(
```
Abhishekkochar, line 102: @weiyuan95 @cuongquangnam this is what's happening with the current implementation.

packages/smartcontracts/src/tests/BridgeProxy.proxyBehaviour.spec.ts
```diff
@@ -0,0 +1,160 @@
...
+      communityAddress,
+      30,
+      communityAddress,
+      2,
+      // Changing version number to 2
+      2,
+    ]);
+
+    // upgrade to the new contract
+    await proxyBridge.upgradeToAndCall(bridgeV1Contract.address, encodedData);
```
Abhishekkochar, line 156: if we to change the init later down, this should be the expected behaviour.

packages/smartcontracts/src/contracts/test/TestBridgeReinitializerTwo.sol
```diff
@@ -0,0 +1,82 @@
...
+    function initialize(
+        address _initialAdmin,
+        address _initialOperational,
+        address _relayerAddress,
+        address _communityWallet,
+        uint256 _fee,
+        address _flushReceiveAddress,
+        uint256 _acceptableRemainingDays,
+        uint8 _version
+    ) external reinitializer(_version) {
```
Abhishekkochar, line 70: And lastly, this is the recommended solution if we want to init the contracts again

Abhishekkochar:
Closing this PR. Not needed anymore.

PR #400, opened 2023-02-15
feature(ui-ux): display browser reload modal

JoshuaC817:
Display a native reload warning modal when there is an unconfirmed transaction ongoing

Fixes #

#### Additional comments?:
Reference for [beforeunload](https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeunload_event) method used
Reasons for not being able to customize a modal or message is that:
1. User should have full control and be able to close any browser
2. Having a customized message will possibly let scammers to mislead users

How it looks like on:
Chrome (Mac)
image

Safari
image

Firefox (Mac)
image

#### Developer Checklist:

- [ ] Unit tests\*
- [ ] Added e2e tests\*
- [ ] Added translations\*


apps/web/src/pages/index.tsx
```diff
@@ -3,9 +3,30 @@ import WelcomeHeader from "@components/WelcomeHeader";
...
+import { UnconfirmedTxnI } from "../types";
 
 function Home() {
   const { isMd } = useResponsive();
+  const { TXN_KEY } = useBridgeFormStorageKeys();
+
+  useEffect(() => {
+    const unloadCallback = (e) => {
+      const event = e;
+      const localData = getStorageItem<UnconfirmedTxnI>(TXN_KEY);
```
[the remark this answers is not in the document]
JoshuaC817: > We might have to change this once we start storing the `txn hash` in the local storage. It's a better indicator because if there's a txn hash -> there's really an unconfirmed txn. This `TXN_KEY` is also being used to save the form state

Done, updated to use `txn hash`



> Another one we might have to consider is if we'll have to reset the form if the user chooses to reload (Form has values,but no unconfirmed txn yet)

Don't think we need to reset the form if the form has value but no unconfirmed txn, this would defeat the purpose of storing it and repopulating the value back in form

JoshuaC817:
EVM > DFC
Will only show reload warning modal when there is `UNCONFIRMED_TXN_HASH_KEY` (which is while waiting for 65 blocks of confirmation)

DFC > EVM
Not showing any reload warning modal since we don't have any long waiting API

apps/web/src/pages/index.tsx
```diff
@@ -1,14 +1,34 @@
...
 function Home() {
   const { ethTxnStatus, isApiSuccess } = useWatchEthTxn();
   const { txnHash, setTxnHash } = useTransactionHashContext();
+  const { UNCONFIRMED_TXN_HASH_KEY } = useBridgeFormStorageKeys();
+
+  useEffect(() => {
+    const unloadCallback = (e) => {
+      const event = e;
+      const unconfirmedHash = getStorageItem<string>(UNCONFIRMED_TXN_HASH_KEY);
+      if (unconfirmedHash !== undefined) {
```
pierregee, line 21:
```suggestion
      const unconfirmedHash = getStorageItem<string>(UNCONFIRMED_TXN_HASH_KEY);
      const unsentFundHash = getStorageItem<string>(UNSENT_FUND_TXN_HASH_KEY);
      if (unconfirmedHash !== undefined || unsentFundHash !== undefined) {
```
Let's also add a warning for this state

pierregee, line 21:
```suggestion
      const unconfirmedHash = getStorageItem<string>(UNCONFIRMED_TXN_HASH_KEY);
      const unsentFundHash = getStorageItem<string>(UNSENT_FUND_TXN_HASH_KEY);
      if (unconfirmedHash !== undefined || unsentFundHash !== undefined) {
```
Let's also add a warning for this state

apps/web/src/pages/index.tsx
```diff
@@ -1,14 +1,34 @@
...
+        // display native reload warning modal if there is unconfirmed txn ongoing
+        event.preventDefault();
+        event.returnValue = "";
+        return "";
+      }
+      return false;
+    };
+    window.addEventListener("beforeunload", unloadCallback);
+    return () => window.removeEventListener("beforeunload", unloadCallback);
+  }, [UNCONFIRMED_TXN_HASH_KEY]);
```
pierregee, line 31:
```suggestion
  }, [UNCONFIRMED_TXN_HASH_KEY, UNSENT_FUND_TXN_HASH_KEY]);
```

pierregee (changes requested):

apps/web/src/pages/index.tsx
JoshuaC817: Alright done!

apps/web/src/pages/index.tsx
JoshuaC817: Done done

pierregee (approved):

Issue #401, opened 2023-02-15
chore(contracts): Contracts should only deal with ETH instead of wETH


weiyuan95:
#### What would you like to be added:
As per title. Contract should only need to deal with native `ETH` tokens instead of `wETH`. This is reverting the logic that https://github.com/WavesHQ/bridge/pull/289 introduced, and only use `ETH` instead.

Note that we do not need to worry about `wETH` in any capacity anymore, we can just deal with `ETH` directly.

#### Why is this needed:
Requirements changes.

PR #406, opened 2023-02-15
feat(Contracts): Removing allowance/daily limit 

### LINKED ISSUE #402: chore(contracts): Remove concept of daily limits
As per title. The contract does not need to concern itself with the concept of a daily limit anymore. This will lead to the removal of most time-based code, such as `changeDailyAllowance`, `allowanceResetTimestamp`, `_acceptableRemainingDays`, etc.

The contract should also now have a per-token hard-limit that will be the basis of comparison when determining the amount of funds that need to be flushed.

Now, the contract should only be interacted with from the UI/server, and will trust the calls from those actors.

Abhishekkochar:
As per the title removing the daily allowance limit from the contract
#402

Fixes #402 

#### Developer Checklist:

- [ ] Unit tests\*
- [ ] Added e2e tests\*
- [ ] Added translations\*


packages/smartcontracts/src/contracts/BridgeV1.sol
```diff
@@ -88,8 +68,8 @@ contract BridgeV1 is UUPSUpgradeable, EIP712Upgradeable, AccessControlUpgradeabl
     bytes32 constant DATA_TYPE_HASH =
         keccak256('CLAIM(address to,uint256 amount,uint256 nonce,uint256 deadline,address tokenAddress)');
 
-    // Mapping to track token address to TokenAllowance
-    mapping(address => TokenAllowance) public tokenAllowances;
+    // Mapping to track the maximum allocation per token address.
```
weiyuan95, line 71:
```suggestion
    // Mapping to track the maximum amount of tokens per token address the contract can hold.
```

packages/smartcontracts/src/contracts/BridgeV1.sol
```diff
@@ -132,30 +109,16 @@ contract BridgeV1 is UUPSUpgradeable, EIP712Upgradeable, AccessControlUpgradeabl
     /**
      * @notice Emitted when a new token is being added to the supported list by only Admin accounts
      * @param supportedToken Address of the token being added to the supported list
-     * @param dailyAllowance Daily allowance of the token
+     * @param tokenCap Maximum allocation per supported token
```
weiyuan95, line 112:
```suggestion
     * @param tokenCap Maximum balance per supported token
```

packages/smartcontracts/src/contracts/BridgeV1.sol
```diff
@@ -132,30 +109,16 @@ contract BridgeV1 is UUPSUpgradeable, EIP712Upgradeable, AccessControlUpgradeabl
...
-     * @param previousTimeStamp The old reset timeStamp of the supported token that is being replaced
-     * @param newTimeStamp The new reset timeStamp of when the supported token starts to be supported
-     */
-    event CHANGE_DAILY_ALLOWANCE(
-        address indexed supportedToken,
-        uint256 indexed changeDailyAllowance,
-        uint256 indexed previousTimeStamp,
-        uint256 newTimeStamp
-    );
-
```
weiyuan95, line 158: Hmm, we should still have a `CHANGE_TOKEN_CAP`? 🤔

packages/smartcontracts/src/contracts/BridgeV1.sol
```diff
@@ -305,24 +244,15 @@ contract BridgeV1 is UUPSUpgradeable, EIP712Upgradeable, AccessControlUpgradeabl
     /**
      * @notice Used by addresses with Admin and Operational roles to add a new supported token and daily allowance
      * @param _tokenAddress The token address to be added to supported list
-     * @param _dailyAllowance Daily allowance set for the token
-     * @param _startAllowanceTimeFrom TimeStamp of when token will be supported.
+     * @param _currentCap To do: not sure about this
```
weiyuan95, line 247: Hmm, what is this referring to?

packages/smartcontracts/src/tests/BridgeProxy.evmToDefi.spec.ts
```diff
@@ -73,116 +45,28 @@ describe('EVM --> DeFiChain', () => {
...
+      const expectedTimestamp = blockBefore.timestamp + 1;
       // Checking contract balance: should be 10 - txFee test tokens
       const txFee = await proxyBridge.transactionFee();
-      const expectedBalance = amountAfterFee({ amount: toWei('10'), transactionFee: txFee });
+      const amount = toWei('10');
+      const expectedBalance = amountAfterFee({ amount, transactionFee: txFee });
+      // Bridging 10 tokens
+      await expect(proxyBridge.bridgeToDeFiChain(ethers.constants.AddressZero, testToken.address, amount))
+        .to.emit(proxyBridge, 'BRIDGE_TO_DEFI_CHAIN')
+        .withArgs(ethers.constants.AddressZero, testToken.address, expectedBalance, expectedTimestamp);
```
weiyuan95, line 65:
```suggestion
        .withArgs(ethers.constants.AddressZero, testToken.address, expectedBalance, time.latest());
```

I think this should be enough, actually 🤔

Abhishekkochar: `await time.latest()` is still one block behind

packages/smartcontracts/src/contracts/BridgeV1.sol
```diff
@@ -244,11 +252,10 @@ contract BridgeV1 is UUPSUpgradeable, EIP712Upgradeable, AccessControlUpgradeabl
     /**
      * @notice Used by addresses with Admin and Operational roles to add a new supported token and daily allowance
      * @param _tokenAddress The token address to be added to supported list
-     * @param _currentCap To do: not sure about this
+     * @param _currentCap maximum balance of tokens the contract can hold per `_tokenAddress`
      */
     function addSupportedTokens(address _tokenAddress, uint256 _currentCap) external {
         if (!checkRoles()) revert NON_AUTHORIZED_ADDRESS();
-        if (_tokenAddress == address(0)) revert ZERO_ADDRESS();
```
weiyuan95, line 318: Hmm, guessing you're removing this for the eventual ETH support?

packages/smartcontracts/src/tests/BridgeProxy.evmToDefi.spec.ts
weiyuan95: ah ok, because you're calling `bridgeToDeFiChain` which adds 1 block. Then maybe `time.latest() + 1`? 🤔

weiyuan95:
We can probably delete the `TestBridge` after this PR as well since we won't need the logic in that contract anymore 🤣

packages/smartcontracts/src/contracts/BridgeV1.sol
Abhishekkochar: Yes, that's correct.

weiyuan95:
Let's also clean up the README.md

Abhishekkochar:
> Let's also clean up the README.md

I will make it part of different PR. Don't want to have too much stuff in this one. Plus, this gives Cuong a base PR for issue #401.

weiyuan95 (approved):
LGTM, will leave it to @cuongquangnam for the final review.

packages/smartcontracts/src/contracts/BridgeV1.sol
```diff
@@ -450,6 +345,18 @@ contract BridgeV1 is UUPSUpgradeable, EIP712Upgradeable, AccessControlUpgradeabl
         emit TRANSACTION_FEE_ADDRESS_CHANGED(oldAddress, _newAddress);
```
cuongquangnam, line 345: cannot comment on this line in the PR: https://github.com/WavesHQ/bridge/blob/4b97ea466ed1b80ab15a85b43527f07b339a4645/packages/smartcontracts/src/contracts/BridgeV1.sol#L330
However, this should be using checkRoles

cuongquangnam (changes requested):

packages/smartcontracts/src/tests/BridgeProxy.transactionFee.spec.ts
```diff
@@ -7,10 +7,10 @@ import { deployContracts } from './testUtils/deployment';
 describe('Transaction fee tests', () => {
   describe('Tx Fee tests', () => {
     describe('DEFAULT_ADMIN_ROLE', () => {
-      it('Successfully implemented the 0.3% fee', async () => {
+      it('Successfully implemented the 0.1% fee', async () => {
         const { proxyBridge } = await loadFixture(deployContracts);
         // Checking if the implemented fee is 0.3%
-        await expect(await proxyBridge.transactionFee()).to.equal(30);
+        await expect(await proxyBridge.transactionFee()).to.equal(10);
```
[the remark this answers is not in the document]
Abhishekkochar: No, We don't.  This is a mistake

packages/smartcontracts/src/contracts/BridgeV1.sol
Abhishekkochar: Oh yes, not sure how I miss this. Double checking @weiyuan95 do we only want admin or both of the addresses to change this?

weiyuan95: Let's have it for both.

cuongquangnam (approved):

PR #420, opened 2023-02-16
feat(server): signclaim - set deadline to 24hours after creation

lykalabrada:
As titled, currently deadline is set to end of day (`23:59:59`) regardless of what time the claim is created.
Due to this change, we now need to also send back the `deadline` to frontend to prevent any time difference when claiming

Fixes #

#### Developer Checklist:

- [ ] Unit tests\*
- [ ] Added e2e tests\*
- [ ] Added translations\*


apps/server/src/utils/DateUtils.ts
```diff
@@ -0,0 +1,7 @@
+export function getNextDayTimestamp(): number {
+  const duration = 1000 * 60 * 60 * 24; // 24 hours in milliseconds
+  const currentTimestamp = Math.floor(Date.now());
```
weiyuan95, line 3: I don't think `Math.floor` does anything here since `Date.now()` should already be an integer 🤔

PR #428, opened 2023-02-17
feat(EVM Bridge Balance Endpoints): Get bridge balance for various assets for EVM

### LINKED ISSUE #410: feat(server): API endpoints for balance checking
We need to add two endpoints

- [ ] `defichain/balance/{tokenSymbol}`
- [x] `ethereum/balance/{tokenSymbol}`

This will query for the balance of the respective tokens on the respective chain's hot wallet/bridge contract.

#### Why is this needed:
Helps to act as a sanity check for the balance on each chain when executing bridging transactions. Bridging transactions will not be able to execute if the wallet on the respective chain does not have enough funds to carry out the transaction.

greggtan:
Get bridge balance for various assets for EVM (USDC, USDT, WETH).

Fixes #410 partially 

#### Additional comments?:
Will work on DFC next

#### Developer Checklist:

- [x] Unit tests\*
- [x] Added e2e tests\*


apps/server/.env
```diff
@@ -11,3 +11,6 @@ DEFICHAIN_NETWORK="Local"
 # https://goerli.etherscan.io/address/0x93fE70235854e7c97A5db5ddfC6eAAb078e99d3C
 BRIDGE_PROXY_ADDRESS="0x93fE70235854e7c97A5db5ddfC6eAAb078e99d3C"
 DEFICHAIN_PRIVATE_KEY="avoid between cupboard there nerve sugar quote foot broom intact seminar culture much anger hold rival moral silly volcano fog service decline tortoise combine"
+USDT_ADDRESS="0x2b3C3A6a97c8a1eEED8C7240880A592440fF5135"
+USDC_ADDRESS="0x0D9C8723B343A8368BebE0B5E89273fF8D712e3C"
+WETH_ADDRESS="0xB4FBF271143F4FBf7B91A5ded31805e42b2208d6"
```
weiyuan95, line 16: Goreli contracts?

apps/server/src/ethereum/services/EVMTransactionConfirmerService.ts
```diff
@@ -23,6 +23,20 @@ export class EVMTransactionConfirmerService {
     );
   }
 
+  async getBalance(tokenSymbol: string): Promise<string> {
+    const contractABI = ['function balanceOf(address) view returns (uint256)'];
+    if (!['WETH', 'USDC', 'USDT'].includes(tokenSymbol)) {
```
weiyuan95, line 28: I would make this an `enum`, maybe something like `SupportedTokenSymbols`. Then I guess in this case, since we need to get the right values from the config, we need to use the enum in `AppConfig` as well so that we only need to maintain the correct symbol in 1 place. ie.

```ts
this.configService.getOrThrow(`ethereum.contracts.${SupportedTokenSymbols[tokenSymbol]}`)
```

and

```ts
// AppConfig.ts
{
  [SupportedTokenSymbols.USDC]: ....
}
```

weiyuan95: Also, there shouldn't be `wETH`, only `ETH` (so there's a slightly different checking flow here). We also need to support checking `wBTC` as well.

apps/server/test-i9n/ethereum/transactionConfirmer.i9n.ts
```diff
@@ -120,4 +139,12 @@ describe('Bridge Service Integration Tests', () => {
     });
     expect(transactionDbRecord?.status).toStrictEqual('CONFIRMED');
   });
+
+  it('Returns the usdc balance of the bridge after bridging 100 usdc and paying 30 Gwei of gas', async () => {
+    const balance = await testing.inject({
+      method: 'GET',
+      url: `/ethereum/balance/USDC`,
+    });
+    expect(JSON.parse(balance.body)).toStrictEqual(99.7);
```
weiyuan95, line 148: I don't think this is 0.3 is gas, I believe it's the transaction fee that is being transferred to the community wallet? The gas cost doesn't affect the amount that is on the bridge since that's on the EVM tx level instead of the wrapped tokens.

apps/server/test-i9n/ethereum/transactionConfirmer.i9n.ts
```diff
@@ -72,12 +83,20 @@ describe('Bridge Service Integration Tests', () => {
     expect(JSON.parse(txReceipt.body).statusCode).toBe(400);
   });
 
+  it('Returns the starting usdc balance of the bridge (should be 0)', async () => {
```
weiyuan95, line 86: We should improve the test coverage to support the other tokens as well. Let's do that in a separate PR.

weiyuan95 (changes requested):

weiyuan95 (changes requested):
One change - we need to support ETH as well, and not just the 3 wrapped tokens.

LGTM with the contract addresses being our mock usdt/usdc/wbtc contracts that we've deployed on testnet.

Will need to improve test coverage (ie. checking that we can actually get the balance for all 3 tokens) in another PR.

apps/server/src/ethereum/services/EVMTransactionConfirmerService.ts
```diff
@@ -23,12 +24,38 @@ export class EVMTransactionConfirmerService {
     );
   }
 
+  async getBalance(tokenSymbol: SupportedTokenSymbols): Promise<string> {
+    const contractABI = ['function balanceOf(address) view returns (uint256)'];
+    if (!SupportedTokenSymbols[tokenSymbol]) {
+      throw new BadRequestException(`Token: "${tokenSymbol}" is not supported`);
+    }
+
+    if (tokenSymbol === 'ETH') {
```
weiyuan95, line 33:
```suggestion
    if (tokenSymbol === SupportedTokenSymbols.ETH) {
```

weiyuan95 (changes requested):
One last quick change!

weiyuan95 (approved):

PR #430, opened 2023-02-17
feature(ui-ux): status api integration

chloezxyy:
- Integrate the endpoint `/bridge/status` to check whether if the bridge is up or not

Fixes #

#### Additional comments?:
- Additional usage of `ApiProvider` wrapper in `Base.tsx` to get `/bridge/status` endpoint to check for condition whether`isUp` is returning true (Bridge is up, don't display maintenance screen)

#### Developer Checklist:

- [ ] Unit tests\*
- [x] Added e2e tests\*
- [ ] Added translations\*


apps/web/src/store/website.tsx
```diff
@@ -72,13 +72,35 @@ export const bridgeApi = createApi({
   }),
 });
 
+export const statusWebsiteSlice = createApi({
```
thedoublejay, line 75: don't add it here, need to create a new flag for it. Maybe change the name to `status.tsx`

thedoublejay (changes requested):

thedoublejay (changes requested):
Also, can you add a simple e2e on it?

1. Intercept the result and change it to true - check if the maintenance text is displayed
2. If not, check if the actual page is displayed

apps/web/src/store/website.tsx
chloezxyy: Refactored a lil

chloezxyy:
> Also, can you add a simple e2e on it?
> 
> 1. Intercept the result and change it to true - check if the maintenance text is displayed
> 2. If not, check if the actual page is displayed

Added!

PR #433, opened 2023-02-17
feature(server): implemented allocateDFCFund endpoint 

fullstackninja864:

Fixes #

#### Developer Checklist:

- [ ] Unit tests\*
- [ ] Added e2e tests\*
- [ ] Added translations\*


apps/server/src/ethereum/services/EVMTransactionConfirmerService.ts
```diff
@@ -121,6 +130,80 @@ export class EVMTransactionConfirmerService {
...
+      const onChainTxnDetail = await this.ethersRpcProvider.getTransaction(transactionHash);
+      const { params } = decodeTxnData(onChainTxnDetail);
+      const { _defiAddress: defiAddress, _tokenAddress: tokenAddress, _amount: amount } = params;
+      const address = ethers.utils.toUtf8String(defiAddress);
+      const evmTokenContract = new ethers.Contract(tokenAddress, ERC20__factory.abi, this.ethersRpcProvider);
+      const wTokenSymbol = await evmTokenContract.symbol();
+      const wTokenDecimals = await evmTokenContract.decimals();
+      const transferAmount = new BigNumber(amount).dividedBy(new BigNumber(10).pow(wTokenDecimals));
+      const dTokenDetails = getDTokenDetailsByWToken(wTokenSymbol, this.network);
+      const txHash = await this.sendService.send(address, {
```
thedoublejay, line 180: Can we store defichain txHash as well? For auditing purposes

thedoublejay (changes requested):

apps/server/src/ethereum/services/EVMTransactionConfirmerService.ts
pierregee, line 180: Need to handle if the hot wallet doesn't have enough balance

pierregee: Also need an explicit response/error that the funds didn't go through so it can be properly handled in ui
image

pierregee (changes requested):

PR #434, opened 2023-02-17
feat(core): Add dynamic server URL

### LINKED ISSUE #419: `server` - replace logic of getting server URLs
The current architecture of the server contains dedicated network and db per server. Example, if we're on testnet, we will use "https://testnet.api.defichain.com (not the final url, just an example)". Same as we config WhaleContexts. We will need to modify https://github.com/WavesHQ/bridge/blob/main/apps/web/src/store/website.tsx to be able to support this. 

- [ ] Replace logic of getting server URLs depending on selected network on the client. 
- [ ] Example, if network is "Local" use "http://localhost:5741", for others, will fill up once deployed
- [ ] Rename the file to "defichain" instead of website
- [ ] Use a similar concept as `newOceanOptions`

kevzzsk:

Add dynamic switch to API url based on the current network environment selected.

Fixes #419 

#### Developer Checklist:

- [ ] Unit tests\*
- [ ] Added e2e tests\*
- [ ] Added translations\*

### DIFF
--- a/apps/web/src/config/networkUrl.ts
+++ b/apps/web/src/config/networkUrl.ts
@@ -2,11 +2,11 @@ import { EnvironmentNetwork } from "@waveshq/walletkit-core";
 
 // TODO: Replace URLs with real URLs
 const BASE_URLS: { [key in EnvironmentNetwork]: string } = {
-  [EnvironmentNetwork.LocalPlayground]: "http://localhost:5741/defichain",
-  [EnvironmentNetwork.RemotePlayground]: "http://localhost:5741/defichain",
-  [EnvironmentNetwork.TestNet]: "http://localhost:5741/defichain",
-  [EnvironmentNetwork.DevNet]: "http://localhost:5741/defichain",
-  [EnvironmentNetwork.MainNet]: "http://localhost:5741/defichain",
+  [EnvironmentNetwork.LocalPlayground]: "http://localhost:5741",
+  [EnvironmentNetwork.RemotePlayground]: "http://localhost:5741",
+  [EnvironmentNetwork.TestNet]: "http://localhost:5741",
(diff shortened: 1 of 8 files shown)


apps/web/src/store/defichain.tsx
```diff
@@ -39,15 +39,14 @@ export const bridgeApi = createApi({
     verify: builder.query<
       any,
       {
-        network: string;
+        baseUrl?: string;
```
kevzzsk, line 42: not sure if this is the best way without TS screaming.

apps/web/src/config/networkUrl.ts
```diff
@@ -0,0 +1,11 @@
+import { EnvironmentNetwork } from "@waveshq/walletkit-core";
+
+const BASE_URLS: { [key in EnvironmentNetwork]: string } = {
```
thedoublejay, line 3: Can you create a github issue for this to replace, then add a TODO comment so we don't forget

apps/web/src/store/defichain.tsx
```diff
@@ -58,9 +57,8 @@ export const bridgeApi = createApi({
       extraOptions: { maxRetries: 3 },
     }),
     getAddressDetail: builder.mutation<AddressDetails, any>({
-      query: ({ network, address }) => ({
-        url: `/wallet/address/${address}`,
-        params: { network },
+      query: ({ baseUrl, address }) => ({
+        url: `${baseUrl}/wallet/address/${address}`,
```
thedoublejay, line 61: I think this will break because we needt o have `defichain` as part of it

thedoublejay (changes requested):

apps/web/src/config/networkUrl.ts
kevzzsk: Done #435

PR #440, opened 2023-02-17
fix(core): multiple poll on consecutive transaction triggered

JoshuaC817:
This PR fixes multiple poll API `/handleTransaction` being triggered
When a transaction is confirmed, do not refresh/leave, proceed with new transaction
Should only have 1 `/handleTransaction` API triggered for new transaction

Fixes #

#### Developer Checklist:

- [ ] Unit tests\*
- [ ] Added e2e tests\*
- [ ] Added translations\*

### DIFF
--- a/apps/web/src/hooks/useWatchEthTxn.ts
+++ b/apps/web/src/hooks/useWatchEthTxn.ts
@@ -13,13 +13,12 @@ export default function useWatchEthTxn() {
 
   const [confirmEthTxn] = useConfirmEthTxnMutation();
 
-  const [isApiLoading, setIsApiLoading] = useState(true);
   const [isApiSuccess, setIsApiSuccess] = useState(false);
   const [ethTxnStatus, setEthTxnStatus] = useState<{
     isConfirmed: boolean;
     numberOfConfirmations: string;
   }>({ isConfirmed: false, numberOfConfirmations: "0" });
-  var pollInterval;
+  let pollInterval;
 
   /* Poll to check if the txn is already confirmed */
   useEffect(() => {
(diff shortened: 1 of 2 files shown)


apps/web/src/hooks/useWatchEthTxn.ts
```diff
@@ -19,23 +19,24 @@ export default function useWatchEthTxn() {
     isConfirmed: boolean;
     numberOfConfirmations: string;
   }>({ isConfirmed: false, numberOfConfirmations: "0" });
+  var pollInterval;
```
thedoublejay, line 22: var

thedoublejay (changes requested):

apps/web/src/hooks/useWatchEthTxn.ts
```diff
@@ -52,9 +53,20 @@ export default function useWatchEthTxn() {
         }
         setIsApiLoading(false);
       }
-      setTimeout(pollConfirmEthTxn, 20000);
     };
-    pollConfirmEthTxn();
+
+    if (pollInterval !== undefined) {
```
thedoublejay, line 58: This can be a function to reuse

thedoublejay (changes requested):

pierregee (approved):
Can check why this is throwing a warning (not urgent). Tested in UI, still ok, just wanna be sure if it has impact @JoshuaC817 
image

apps/web/src/hooks/useWatchEthTxn.ts
pierregee: We'll handle it in another PR. No impact on UI, tested it

pierregee:
Added a call to api on load

PR #452, opened 2023-02-20
fix(ui-ux): fix race condition when calling SC's write function

lykalabrada:
Few fixes for EVM -> DFC transfer flow:
- Disable "Confirm Transfer" button when SC's write functions are not yet ready, this will prevent the race condition right now

Fixes #

#### Developer Checklist:

- [ ] Unit tests\*
- [ ] Added e2e tests\*
- [ ] Added translations\*


pierregee (changes requested):
image
The Bridge UI modal is stuck even after rejecting from connector (metamask)

EDIT: happens on main too!

pierregee (approved):

PR #460, opened 2023-02-21
chore(e2e): added integration test for allocate fund

fullstackninja864:

Fixes #

#### Developer Checklist:

- [ ] Unit tests\*
- [ ] Added e2e tests\*
- [ ] Added translations\*


apps/server/src/ethereum/services/EVMTransactionConfirmerService.ts
```diff
@@ -193,6 +179,20 @@ export class EVMTransactionConfirmerService {
         throw new Error('Transaction is not yet confirmed');
       }
 
+      const txReceipt = await this.ethersRpcProvider.getTransactionReceipt(transactionHash);
```
pierregee, line 182: txReceipt can be null. This happens when transaction is still pending. See `handleTransaction` service

pierregee (approved):
LGTM. one comment, but not a blocker

apps/server/src/ethereum/services/EVMTransactionConfirmerService.ts
fullstackninja864: added condition to check that

pierregee (approved):

PR #464, opened 2023-02-21
feat(Contract): unable to claim tokens from DFI to EVM

### LINKED ISSUE #461: Signing a claim with version 2.0 is causing an error on client side claiming process
#### What happened:
Signing and claiming of tokens is not working properly. On the server side, we are getting the latest version from the contract which is `version 2.0` and uses that to sign the claim, but then when user tries to claim the tokens, it is failing.

User should be able to claim the tokens signed by the server using version 2.0

When using `version 1.0` to sign the claims, there is no issue claiming the token on the client side.

Abhishekkochar:
 `version` should be `2.0` after the upgrade. However, tx is failing with this one.
 Tx will only proceed if the `version` is `1.0`.

Fixes #461 

#### Developer Checklist:

- [ ] Unit tests\*
- [ ] Added e2e tests\*
- [ ] Added translations\*


packages/smartcontracts/src/tests/BridgeProxy.proxyBehaviour.spec.ts
```diff
@@ -0,0 +1,84 @@
...
+    ({ proxyBridge, defaultAdminSigner, operationalAdminSigner, communityAddress, testToken } = await loadFixture(
+      deployContracts,
+    ));
+    const BridgeUpgradeable = await ethers.getContractFactory('BridgeV2');
+    const bridgeUpgradeable = await BridgeUpgradeable.deploy();
+    await bridgeUpgradeable.deployed();
+    await proxyBridge.upgradeTo(bridgeUpgradeable.address);
+    // `version` should be `2.0`. However, tx is failing with this one.
+    // Tx will only proceed if the `version` is `1.0` after the upgrade
+    domainData = {
```
Abhishekkochar, line 37: @cuongquangnam @weiyuan95  here

packages/smartcontracts/src/contracts/test/reinitialize/BridgeV2.sol
```diff
@@ -0,0 +1,451 @@
...
+     * @param _fee Fee charged on each transcation (initial fee: 0.3%)
+     */
+    function initialize(
+        address _initialAdmin,
+        address _initialOperational,
+        address _relayerAddress,
+        address _communityWallet,
+        uint256 _fee,
+        address _flushReceiveAddress,
+        uint8 _version
```
weiyuan95, line 216: Should this be set as a constant to avoid any avoidance of doubt? Or will that mess up the storage?

packages/smartcontracts/src/tests/BridgeProxy.proxyBehaviour.spec.ts
```diff
@@ -0,0 +1,242 @@
...
+      // flushReceiveAddress
+      defaultAdminSigner.address,
+    ]);
+    // Init the contract with `initializer` modifier
+    await expect(proxyBridge.upgradeToAndCall(bridgeUpgradeableInit.address, encodedData)).to.be.revertedWith(
+      'Initializable: contract is already initialized',
+    );
+  });
+
+  it('UnSuccessfully reverting when claiming with new contract version', async () => {
```
weiyuan95, line 161:
```suggestion
  it('should revert when claiming with new contract version', async () => {
```

weiyuan95 (commented):
Few quick comments, if not LGTM!

packages/smartcontracts/src/contracts/test/reinitialize/BridgeV2.sol
weiyuan95: Resolving - doing string (the current type on the contract) to int (the necessary type for `reinitializer`) is going to cause some overheads which we don't need to deal with right now

weiyuan95 (approved):
Approving this first, @cuongquangnam can look at it tomorrow to catch anything i might have missed.

Next steps would be to clean up the project directory, since having `contracts/test/reinitialize/BridgeV2` is confusing when there's a `contracts/BridgeV1`.

Abhishekkochar:
> Approving this first, @cuongquangnam can look at it tomorrow to catch anything i might have missed.
> 
> Next steps would be to clean up the project directory, since having `contracts/test/reinitialize/BridgeV2` is confusing when there's a `contracts/BridgeV1`.

We possibly rename it to BridgeV2Testnet

PR #467, opened 2023-02-21
feat(smartcontracts): Fix bug in claimFund and add several security improvements

### LINKED ISSUE #456: Use `safeTransfer` instead of `transfer`
Use `safeTransfer` instead of 'transfer'
https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC20/utils/SafeERC20.sol

#### Why is this needed:
`transfer()` fn is being used to transfer balance from users to contract. While this works perfectly fine. However there are tokens that returns bool(false) when the tx hasn't transferred the token to contract or vice versa, instead of reverting. 

This can result in emitting a bridge event without actual transferred of tokens.

cuongquangnam:
1. Disable initialization for implementation smart contract
2. Add SafeERC20 utility libraries by OpenZeppelin to accommodate different variants of ERC20 tokens
3. Fix a bug in claimFund which sent ETH to msg.sender instead of sending fund to _to address

Fixes #456

#### Developer Checklist:

- [ ] Unit tests\*
- [ ] Added e2e tests\*
- [ ] Added translations\*


packages/smartcontracts/src/contracts/BridgeV1.sol
```diff
@@ -1,14 +1,11 @@
 // SPDX-License-Identifier: MIT
 pragma solidity ^0.8.13;
```
weiyuan95, line 2: These changes should be made on the `BridgeV2` contract first right? @Abhishekkochar

packages/smartcontracts/src/contracts/BridgeV1.sol
```diff
@@ -70,29 +67,31 @@ error AMOUNT_PARAMETER_NOT_ZERO_WHEN_BRIDGING_ETH();
...
+    string public constant VERSION = '1';
 
     address public constant ETH = address(0);
 
     // Mapping to track the address's nonce
     mapping(address => uint256) public eoaAddressToNonce;
 
     // Enumerable set of supportedTokens
-    EnumerableSet.AddressSet internal supportedTokens;
+    EnumerableSetUpgradeable.AddressSet internal supportedTokens;
```
weiyuan95, line 89: I'm assuming we're using this because of https://docs.openzeppelin.com/contracts/4.x/upgradeable

Can we double check if we've missed out any other contracts that should be `*Upgradable`?

packages/smartcontracts/src/contracts/BridgeV1.sol
Abhishekkochar: Ideally yes, we can make this part of V3.

packages/smartcontracts/src/tests/BridgeProxy.defiToEvm.spec.ts
```diff
@@ -65,6 +65,30 @@ describe('DeFiChain --> EVM', () => {
...
+        amount: toWei('10'),
+        nonce: 0,
+        deadline: ethers.constants.MaxUint256,
+        tokenAddress: testToken.address,
+      };
+
+      const signature = await defaultAdminSigner._signTypedData(domainData, eip712Types, eip712Data);
+      // Checking Balance before claiming fund, should be 0
+      expect(await testToken.balanceOf(operationalAdminSigner.address)).to.equal(0);
+      await proxyBridge.claimFund(
```
Abhishekkochar, line 80:
```suggestion
      await proxyBridge.connect(defaultAdminSigner).claimFund(
```

packages/smartcontracts/src/tests/BridgeProxy.defiToEvm.spec.ts
```diff
@@ -272,5 +273,163 @@ describe('DeFiChain --> EVM', () => {
...
+        nonce: 0,
+        deadline: ethers.constants.MaxUint256,
+        tokenAddress: ethers.constants.AddressZero,
+      };
+      // Relayer address is defaultAdminSigner, if not signed by relayer address, txn should fail.
+      const signature = await operationalAdminSigner._signTypedData(domainData, eip712Types, eip712Data);
+      const balanceOperationalAdminBeforeClaim = await ethers.provider.getBalance(operationalAdminSigner.address);
+      await expect(
+        proxyBridge.claimFund(
+          operationalAdminSigner.address,
```
Abhishekkochar, line 322: We are already testing this case above. Can we change this to `arbitrarySigner`? Will still revert with custom error, however more diversify.

Abhishekkochar (changes requested):
Minor changes. Other than this LGTM.

packages/smartcontracts/src/contracts/BridgeV1.sol
cuongquangnam: Yeah, I think it is a good practice that all contracts imported for the implementation contract are upgradeable (though we don't need to use upgradeable version for the smart contracts that don't have state variables IMO) Also, this helps avoid duplicate of deployed smart contracts (e.g. in the previous [deployment](https://goerli.etherscan.io/address/0x0d269d012fffd93b448c0090740241c8e9556031#code) there are Strings and StringsUpgrdeable contracts)

Abhishekkochar: We will be using `StringsUpgrdeable` if we have to reinit the contract. [https://github.com/WavesHQ/bridge/blob/54dc0b7fb34b66637486c18a929a1695cc2b9f20/pa[…]rtcontracts/src/contracts/test/reinitialize/BridgeV2TestNet.sol](https://github.com/WavesHQ/bridge/blob/54dc0b7fb34b66637486c18a929a1695cc2b9f20/packages/smartcontracts/src/contracts/test/reinitialize/BridgeV2TestNet.sol#L408) unless backend does the conversion from big number to string.

packages/smartcontracts/src/contracts/test/reinitialize/BridgeV2TestNet.sol
```diff
@@ -70,22 +67,23 @@ error AMOUNT_PARAMETER_NOT_ZERO_WHEN_BRIDGING_ETH();
 error MSG_VALUE_NOT_ZERO_WHEN_BRIDGING_ERC20();
 
 contract BridgeV2TestNet is UUPSUpgradeable, EIP712Upgradeable, AccessControlUpgradeable {
-    using EnumerableSet for EnumerableSet.AddressSet;
+    using EnumerableSetUpgradeable for EnumerableSetUpgradeable.AddressSet;
```
Abhishekkochar, line 70: Oh forgot about these one  😅😅

packages/smartcontracts/src/tests/BridgeProxy.defiToEvm.spec.ts
weiyuan95: We are also testing for ETH here (ie. `0x00`) so I would say it's okay.

apps/server/src/ethereum/services/EVMTransactionConfirmerService.ts
```diff
@@ -4,8 +4,8 @@ import { ConfigService } from '@nestjs/config';
 import { EthereumTransactionStatus } from '@prisma/client';
 import { EnvironmentNetwork } from '@waveshq/walletkit-core';
 import BigNumber from 'bignumber.js';
-import { BigNumber as EthBigNumber, Contract, ethers } from 'ethers';
-import { BridgeV2TestNet__factory, ERC20__factory } from 'smartcontracts';
+import { BigNumber as EthBigNumber, ethers } from 'ethers';
+import { BridgeV1, BridgeV1__factory, ERC20__factory } from 'smartcontracts';
```
Abhishekkochar, line 8: I'm certain we need `BridgeV2TestNet__factory` in the backend for testnet. Abi is different from BridgeV1__factory. @lykalabrada am I right?

lykalabrada: Previously we had to use `BridgeV2TestNet__factory` because of [this issue](https://github.com/WavesHQ/bridge/issues/461) with `BridgeV1__factory`.

But I just tried switching back to `BridgeV1__factory` and `claimFund` is now working as well.
By right, should they be the same? I think BridgeV1__factory is more similar to our [ABI on testnet](https://goerli.etherscan.io/address/0x8b929b940bc0d4d26072e0d8d659b08961d6ff09#code)

Abhishekkochar (approved):

apps/server/src/ethereum/services/EVMTransactionConfirmerService.ts
weiyuan95: Bridgev1 and bridgev2tesnet contracts are at parity logic wise, we just need to use bridgev1 for the `initializer` modifer

PR #478, opened 2023-02-22
feat(DFC Balance): added service and integration tests to get asset balance of hotwallet

### LINKED ISSUE #410: feat(server): API endpoints for balance checking

greggtan:
Gets balance of all assets of DFC hotwallet (USDC, USDT, ETH, BTC, DFI). 

Fixes #410

#### Developer Checklist:

- [x] Unit tests\*
- [x] Added e2e tests\*

### DIFF
--- a/apps/server/test-i9n/defichain/balance.i9n.ts
+++ b/apps/server/test-i9n/defichain/balance.i9n.ts
@@ -124,14 +124,17 @@ describe('DeFiChain Send Transaction Testing', () => {
     );
     await defichain.generateBlock();
 
-    // // Send 10 DFI to hotwallet
+    // Send 10 UTXO to hotwallet
     await defichain.playgroundRpcClient?.wallet.sendToAddress(hotWalletAddress, 10);
     await defichain.generateBlock();
 
     expect(await getBalance('BTC')).toStrictEqual(10);
     expect(await getBalance('USDC')).toStrictEqual(10);
     expect(await getBalance('USDT')).toStrictEqual(10);
+
+    // Delay to workaround throttler exception
     await delay(30000);
+
     expect(await getBalance('ETH')).toStrictEqual(10);
     expect(await getBalance('DFI')).toStrictEqual(10);
   });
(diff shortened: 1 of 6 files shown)


apps/server/test-i9n/defichain/balance.i9n.ts
```diff
@@ -0,0 +1,138 @@
...
+        {},
+        {
+          [hotWalletAddress]: `10@ETH`,
+        },
+      ],
+      'number',
+    );
+    await defichain.generateBlock();
+
+    // // Send 10 DFI to hotwallet
```
thedoublejay, line 127:
```suggestion
    // // Send 10 DFI UTXO to hotwallet
```

apps/server/test-i9n/defichain/balance.i9n.ts
```diff
@@ -0,0 +1,138 @@
...
+    await defichain.playgroundRpcClient?.wallet.sendToAddress(hotWalletAddress, 10);
+    await defichain.generateBlock();
+
+    expect(await getBalance('BTC')).toStrictEqual(10);
+    expect(await getBalance('USDC')).toStrictEqual(10);
+    expect(await getBalance('USDT')).toStrictEqual(10);
+    await delay(30000);
```
thedoublejay, line 134: Curious why we need a delay here

thedoublejay (changes requested):

apps/server/test-i9n/defichain/balance.i9n.ts
greggtan: throttle limit throwing error: {"message": "ThrottlerException: Too Many Requests", "statusCode": 429} , so I implemented a delay to workaround it!

pierregee: Ya encountered this too, not sure what's the right way, so I also had to resort to [this](https://github.com/WavesHQ/bridge/blob/main/apps/server/test-i9n/defichain/verify.i9n.ts#L20)! 🙃

PR #539, opened 2023-02-24
fix(core): handle address details global state

### LINKED ISSUE #519: UX - new address generated without user knowing
#### What UI/UX improvements would you like to be implemented:
– Inform user on step **2 Transact** that qr code address will re-generated upon selecting `ⓧ`

#### Why is this needed:
**Actual:**
– If user sends funds to qr code generated address, and accidentally selects `ⓧ `instead of `Verify transfer`, user may not notice that by selecting `Review Transaction` again generates a new address

#### Which component/screens does this concern?:
DFC → ETH:
on Review Transaction pop-up modal, step **2 Transact**

**To reproduce:**
1. User select DFC → ETH > at Amount to transfer, input `0.0001` dBTC > at **Address** field, paste `MetaMask address` > `Review transaction`
2. at **Review transaction** pop-up modal, `note:` qr code address
3. Send `0.0001` dBTC to address, select `ⓧ` > redirects to Quantum with previous inputs
4. at Quantum, user selects `Review transaction` > redirects to **Review transaction** pop-up modal >` note:` a new address is generated, but user might think that tx was being processed and selects Verify transfer

pierregee:

- [x] Disable refund address once user reaches "Step 2" wherein the DFC address is already generated with refundAddress tied in to it
- [x] Move the addressDetails in context so when a user do a "Retry transfer", a new refund address is "restored". Previously, it only gets restored if user refreshes (gets from local storage)
- [x] Move the form state in context so user can reset the form anytime. Previously the user have to refresh the page to see the "Reset form"

Fixes #519 

#### Developer Checklist:

- [ ] Unit tests\*
- [ ] Added e2e tests\*
- [ ] Added translations\*


lykalabrada (approved):
Looks good. Just need to address failing tests

PR #561, opened 2023-02-24
refactor(ui-ux): add qr address to step 3 of dfc > evm

JoshuaC817:
To prevent users from needing to go back to step 2, we retain displaying the QR address. 
This is so users will always have access on the address they need to send to.

Fixes #

#### Developer Checklist:

- [x] Your UI implementation visually matched the rendered design\*
- [ ] Unit tests\*
- [ ] Added e2e tests\*
- [ ] Added translations\*


pierregee (commented):
All good with the QR code part, but regarding the try again button, re we hiding/disabling it when it's validating?
image

JoshuaC817:
> All good with the QR code part, but regarding the try again button, re we hiding/disabling it when it's validating? image

checking with designers on this, will update soon

pierregee (approved):

PR #580, opened 2023-02-27
feat(Contract): Post Audit features request

### LINKED ISSUE #579: Post audit features request

- A new flushFund function for each token, and a flushFund function with from, to index in the token address array as parameters
- Tx Fee upper limit
- Withdraw now should withdraw to flushFundAddress, not msg.sender anymore

Abhishekkochar:
Added cap on tx fee,
Should be able to flush fund per token and also by providing `from` and `to` index

Fixes #579

#### Developer Checklist:

- [ ] Unit tests\*
- [ ] Added e2e tests\*
- [ ] Added translations\*


packages/smartcontracts/src/contracts/BridgeV1.sol
```diff
@@ -331,29 +342,57 @@ contract BridgeV1 is UUPSUpgradeable, EIP712Upgradeable, AccessControlUpgradeabl
      * @param amount Requested amount to be withdraw. Amount would be in the denomination of ETH
      */
     function withdraw(address _tokenAddress, uint256 amount) external onlyRole(DEFAULT_ADMIN_ROLE) {
+        address _flushReceiveAddress = flushReceiveAddress;
         if (_tokenAddress == ETH) {
-            (bool sent, ) = msg.sender.call{value: amount}('');
+            (bool sent, ) = _flushReceiveAddress.call{value: amount}('');
```
weiyuan95, line 347: Oh, we can't just use `flushReceiveAddress` (the constant) here? 🤔

packages/smartcontracts/src/contracts/BridgeV1.sol
```diff
@@ -331,29 +342,57 @@ contract BridgeV1 is UUPSUpgradeable, EIP712Upgradeable, AccessControlUpgradeabl
...
-        } else IERC20Upgradeable(_tokenAddress).safeTransfer(msg.sender, amount);
-        emit WITHDRAWAL_BY_OWNER(msg.sender, _tokenAddress, amount);
+        } else IERC20Upgradeable(_tokenAddress).safeTransfer(_flushReceiveAddress, amount);
+        emit WITHDRAWAL_BY_OWNER(_flushReceiveAddress, _tokenAddress, amount);
     }
 
     /**
      * @notice Function to flush the excess funds across supported tokens to a hardcoded address
-     * anyone can call this function
+     * anyone can call this function, if pass (0,3), only token at index 0, 1 and 2 will be flushed.
```
weiyuan95, line 355:
```suggestion
     * anyone can call this function. For example, calling flushMultipleTokenFunds(0,3), only the tokens at index 0, 1 and 2 will be flushed.
```

Let's call the function `flushMultipleTokenFunds`

packages/smartcontracts/src/contracts/BridgeV1.sol
```diff
@@ -331,29 +342,57 @@ contract BridgeV1 is UUPSUpgradeable, EIP712Upgradeable, AccessControlUpgradeabl
...
+     * @param _fromIndex Finishing index for array `supportedTokens.values()` to flush from
```
weiyuan95, line 356:
```suggestion
     * @param _fromIndex End index for array `supportedTokens.values()` to flush from
```

packages/smartcontracts/src/contracts/BridgeV1.sol
```diff
@@ -331,29 +342,57 @@ contract BridgeV1 is UUPSUpgradeable, EIP712Upgradeable, AccessControlUpgradeabl
...
+     * @param _toIndex Starting index for array `supportedTokens.values()` to flush to
      */
-    function flushFund() external {
-        for (uint256 i = 0; i < supportedTokens.length(); ++i) {
+    function flushFund(uint256 _fromIndex, uint256 _toIndex) external {
+        if (_toIndex > supportedTokens.length()) revert INVALID_TOINDEX();
```
weiyuan95, line 360: Does solidity expect indexes less than 0? We should check `_fromIndex >= 0` as well? 🤔

packages/smartcontracts/src/tests/BridgeProxy.transactionFee.spec.ts
```diff
@@ -51,6 +51,13 @@ describe('Transaction fee tests', () => {
         );
       });
     });
+    it('Successfully revert if the fee is greater than `MAX_FEE`', async () => {
+      const { proxyBridge } = await loadFixture(deployContracts);
+      // Admin should successfully changes the tx fees to 101%
```
weiyuan95, line 56:
```suggestion
      // Admin should not be able to change the tx fees to more than 100%
```

packages/smartcontracts/src/tests/BridgeProxy.flushFund.spec.ts
```diff
@@ -46,6 +48,41 @@ describe('Test flushFund functionalities', () => {
     expect(balanceETHAfterFlush.sub(balanceETHBeforeFlush)).to.equal(toWei('40'));
   });
 
+  it('Should flush the funds successfully when there is initial redundant funds per `tokenAddress`', async () => {
```
weiyuan95, line 51:
```suggestion
  it('Should flush the funds per token successfully when there is initial redundant funds per `tokenAddress`', async () => {
```

This is for the per token scenario?

weiyuan95 (changes requested):

packages/smartcontracts/src/contracts/BridgeV1.sol
Abhishekkochar: It's not a constant though. It is a state variable.

packages/smartcontracts/src/tests/BridgeProxy.flushFund.spec.ts
Abhishekkochar: Oh yes, fixed

packages/smartcontracts/src/contracts/BridgeV1.sol
Abhishekkochar: This is an `uint256 _fromIndex` We can't pass index that is less than 0 or negative here to start with.

Abhishekkochar: Screenshot 2023-02-28 at 12 15 06 PM

cuongquangnam (changes requested):

packages/smartcontracts/src/contracts/BridgeV1.sol
```diff
@@ -331,29 +342,57 @@ contract BridgeV1 is UUPSUpgradeable, EIP712Upgradeable, AccessControlUpgradeabl
...
+        emit FLUSH_FUND();
+    }
+
+    /**
+     * @notice Function to flush the excess funds across supported token to a hardcoded address
+     * anyone can call this function
+     * @param _tokenAddress address of the token to be flushed
+     */
+    function flushFundPerToken(address _tokenAddress) external {
+        address _flushReceiveAddress = flushReceiveAddress;
```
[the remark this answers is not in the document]
Abhishekkochar: Actually yes, you are right. Forgot to see that was conditional.

packages/smartcontracts/src/contracts/BridgeV1.sol
```diff
@@ -331,29 +342,57 @@ contract BridgeV1 is UUPSUpgradeable, EIP712Upgradeable, AccessControlUpgradeabl
...
+        }
+        emit FLUSH_FUND();
+    }
+
+    /**
+     * @notice Function to flush the excess funds across supported token to a hardcoded address
+     * anyone can call this function
+     * @param _tokenAddress address of the token to be flushed
+     */
+    function flushFundPerToken(address _tokenAddress) external {
```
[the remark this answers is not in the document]
Abhishekkochar: Good point, However, I don't see the case where we will have the funds of unsupported token.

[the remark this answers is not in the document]
cuongquangnam: I think better have a guard rather than not, because it is open to public, not like withdraw function where only privileged ones can access

packages/smartcontracts/src/contracts/BridgeV1.sol
weiyuan95: Ah, set after init. Got it.

packages/smartcontracts/src/contracts/BridgeV1.sol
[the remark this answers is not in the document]
weiyuan95: Let's have this guard as a sanity check. For withdraw, yes, it's ok to not make the check, but when it comes to flushing the funds we should only need to flush for tokens we support 🤔

packages/smartcontracts/src/contracts/BridgeV1.sol
weiyuan95: Ah, got it. It's `uint` so it cannot be negative.

packages/smartcontracts/src/contracts/BridgeV1.sol
```diff
@@ -331,29 +342,57 @@ contract BridgeV1 is UUPSUpgradeable, EIP712Upgradeable, AccessControlUpgradeabl
...
+        emit WITHDRAWAL_BY_OWNER(_flushReceiveAddress, _tokenAddress, amount);
     }
 
     /**
-     * @notice Function to flush the excess funds across supported tokens to a hardcoded address
-     * anyone can call this function
+     * @notice anyone can call this function. For example, calling flushMultipleTokenFunds(0,3),
+     * only the tokens at index 0, 1 and 2 will be flushed.
+     * @param _fromIndex _fromIndex End index for array `supportedTokens.values()` to flush from
+     * @param _toIndex Starting index for array `supportedTokens.values()` to flush to
```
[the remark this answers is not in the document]
weiyuan95: Good point on the `inclusive` and `exclusive`.

packages/smartcontracts/src/tests/BridgeProxy.flushFund.spec.ts
```diff
@@ -48,6 +48,14 @@ describe('Test flushMultipleTokenFunds functionalities', () => {
     expect(balanceETHAfterFlush.sub(balanceETHBeforeFlush)).to.equal(toWei('40'));
   });
 
+  it('Should revert if ERC20 no supported', async () => {
```
weiyuan95, line 51:
```suggestion
  it('Should revert if ERC20 is not supported', async () => {
```

PR #587, opened 2023-02-27
feat(smart contracts): Add timelock as per audit suggestions

cuongquangnam:
Add timelock contract
My current idea for how this should work:
- We deploy one TimelockController as a DEFAULT_ADMIN_ROLE who can withdraw money from the bridge contract.
- We deploy another TimelockController as a OPERATIONAL_ADMIN_ROLE who can change configuration parameters.

Fixes #

#### Developer Checklist:

- [ ] Unit tests\*
- [ ] Added e2e tests\*
- [ ] Added translations\*


weiyuan95:
For posterity: 

We will be having 2 roles
- `DEFAULT_ADMIN` role, where the address is the TimeLock contract
- `withdraw` role, where the address is the multisig contract

They are mutually exclusive roles, where the withdraw role can _only_ call withdraw funds. Every other state-change method call must go through the TimeLock contract via the `DEFAULT_ADMIN` role.

The TimeLock contract will also have a proposer and executor role that is tied to the multisig contract as well. The minimum delay will be 3 days.

packages/smartcontracts/src/tests/BridgeProxy.proxyBehaviour.spec.ts
```diff
@@ -114,117 +114,118 @@ describe('Proxy behaviour', () => {
...
-        1,
-        ethers.constants.MaxUint256,
-        testToken.address,
-        signatureV2,
-      ),
-    ).to.be.revertedWithCustomError(proxyBridge, 'FAKE_SIGNATURE');
-    // Checking Balance after claiming fund, should be 10
-    expect(await testToken.balanceOf(defaultAdminSigner.address)).to.equal(toWei('10'));
-  });
+  // not necessary tests
```
weiyuan95, line 117: Agree that these are probably not necessary since we know how the initializer works now. Let's just delete instead of commenting it out.

packages/smartcontracts/src/tests/BridgeProxy.Withdraw.spec.ts
```diff
@@ -31,67 +33,67 @@ describe('Withdrawal tests', () => {
...
 
     it('Unable to withdraw more ERC20 than the balance of the Bridge', async () => {
-      const { proxyBridge, testToken, defaultAdminSigner } = await loadFixture(deployContracts);
+      const { proxyBridge, testToken, withdrawSigner } = await loadFixture(deployContracts);
       // Contract balance of testToken is '0'
-      // Test should be revert with a mention string if Admin requesting amount bigger than actual balance of the Bridge.
-      await expect(
-        proxyBridge.connect(defaultAdminSigner).withdraw(testToken.address, toWei('110')),
-      ).to.be.revertedWith('ERC20: transfer amount exceeds balance');
+      // Test should be revert with a mention string if requesting amount bigger than actual balance of the Bridge.
```
weiyuan95, line 59:
```suggestion
      // Test should revert if requesting amount bigger than actual balance of the Bridge.
```

packages/smartcontracts/src/tests/BridgeProxy.Withdraw.spec.ts
```diff
@@ -31,67 +33,67 @@ describe('Withdrawal tests', () => {
...
-      await expect(
-        proxyBridge.connect(operationalAdminSigner).withdraw(testToken.address, toWei('20')),
-      ).to.be.revertedWith(
-        // address from hardcoded Hardhat network accounts
-        'AccessControl: account 0x70997970c51812dc3a010c7d01b50e0d17dc79c8 is missing role 0x0000000000000000000000000000000000000000000000000000000000000000',
+  describe('DEFAULT_ADMIN_ROLE', () => {
+    it('Unsuccessful withdrawal by those with DEFAULT_ADMIN_ROLE but does not have WITHDRAW_ROLE', async () => {
+      const { proxyBridge, testToken, defaultAdminSigner } = await loadFixture(deployContracts);
+      const WITHDRAW_ROLE = ethers.utils.solidityKeccak256(['string'], ['WITHDRAW_ROLE']);
+      // sanity check on whether we need to make the defaultadminsigner.address to lower-case
```
weiyuan95, line 81: Hmm, what's this? Don't see any lower case conversion going on here.

weiyuan95 (changes requested):
LGTM, just some minor comments.

Issue to improve test coverage is tracked in #600

packages/smartcontracts/src/contracts/test/reinitialize/BridgeV2TestNet.sol
```diff
@@ -201,13 +215,20 @@ contract BridgeV2TestNet is UUPSUpgradeable, EIP712Upgradeable, AccessControlUpg
...
      * @param _version Contract's version
      */
     function initialize(uint8 _version) external reinitializer(_version) {
         __EIP712_init(NAME, StringsUpgradeable.toString(_version));
     }
 
+    /**
+     * @notice To get the current version of the contract
+     */
+    function version() external view returns (string memory) {
```
Abhishekkochar, line 228: Please move this view fn() with other view fn(), at the end of contract. Same for BridgeV1. 
This is my bad, I placed  it right after init.

Abhishekkochar (changes requested):
Quick fix, Everything else LGTM.

packages/smartcontracts/src/tests/BridgeProxy.Withdraw.spec.ts
cuongquangnam: Oh, just a sanity check to see whether I need to transform the address to lower case when inputting into the hasRole function like the case of AccessControl error:
``` 
`AccessControl: account ${defaultAdminSigner.address.toLowerCase()} is missing role ${WITHDRAW_ROLE}`
```

weiyuan95: Ah - it's below. Got it.

weiyuan95: I can do it from a suggestion

packages/smartcontracts/src/tests/BridgeProxy.Withdraw.spec.ts
```diff
@@ -31,67 +33,67 @@ describe('Withdrawal tests', () => {
...
+      expect(await proxyBridge.hasRole(`0x${'0'.repeat(64)}`, defaultAdminSigner.address)).to.equal(true);
+      expect(await proxyBridge.hasRole(WITHDRAW_ROLE, defaultAdminSigner.address)).to.equal(false);
+      // Withdrawal by defaultAdminSigner should be rejected if it has not been granted WITHDRAW_ROLE
+      await expect(proxyBridge.connect(defaultAdminSigner).withdraw(testToken.address, toWei('20'))).to.be.revertedWith(
```
weiyuan95, line 85:
```suggestion
      expect(await proxyBridge.hasRole(`0x${'0'.repeat(64)}`, defaultAdminSigner.address)).to.equal(true);
      expect(await proxyBridge.hasRole(WITHDRAW_ROLE, defaultAdminSigner.address)).to.equal(false);
      // Withdrawal by defaultAdminSigner should be rejected if it has not been granted WITHDRAW_ROLE
      // sanity check on whether we need to make the defaultadminsigner.address to lower-case
      await expect(proxyBridge.connect(defaultAdminSigner).withdraw(testToken.address, toWei('20'))).to.be.revertedWith(
```

packages/smartcontracts/src/tests/BridgeProxy.Withdraw.spec.ts
```diff
@@ -31,67 +33,67 @@ describe('Withdrawal tests', () => {
...
+        `AccessControl: account ${defaultAdminSigner.address.toLowerCase()} is missing role ${WITHDRAW_ROLE}`,
       );
     });
   });
   describe('ARBITRARY_EOA', () => {
     it('Unsuccessful withdrawal by other EOA', async () => {
       const { proxyBridge, arbitrarySigner, testToken } = await loadFixture(deployContracts);
-      // Withdrawal by another Admin should be rejected
+      const WITHDRAW_ROLE = ethers.utils.solidityKeccak256(['string'], ['WITHDRAW_ROLE']);
+      expect(await proxyBridge.hasRole(WITHDRAW_ROLE, arbitrarySigner.address)).to.equal(false);
```
weiyuan95, line 94:
```suggestion
      expect(await proxyBridge.hasRole(WITHDRAW_ROLE, arbitrarySigner.address)).to.equal(false);
      // sanity check on whether we need to make the defaultadminsigner.address to lower-case
```

weiyuan95 (changes requested):

packages/smartcontracts/src/tests/BridgeProxy.Withdraw.spec.ts
```diff
@@ -92,6 +92,7 @@ describe('Withdrawal tests', () => {
       const { proxyBridge, arbitrarySigner, testToken } = await loadFixture(deployContracts);
       const WITHDRAW_ROLE = ethers.utils.solidityKeccak256(['string'], ['WITHDRAW_ROLE']);
       expect(await proxyBridge.hasRole(WITHDRAW_ROLE, arbitrarySigner.address)).to.equal(false);
+      // sanity check on whether we need to make the defaultadminsigner.address to lower-case
```
cuongquangnam, line 95: do we need this comment here?

packages/smartcontracts/src/tests/BridgeProxy.Withdraw.spec.ts
```diff
@@ -78,10 +78,10 @@ describe('Withdrawal tests', () => {
     it('Unsuccessful withdrawal by those with DEFAULT_ADMIN_ROLE but does not have WITHDRAW_ROLE', async () => {
       const { proxyBridge, testToken, defaultAdminSigner } = await loadFixture(deployContracts);
       const WITHDRAW_ROLE = ethers.utils.solidityKeccak256(['string'], ['WITHDRAW_ROLE']);
-      // sanity check on whether we need to make the defaultadminsigner.address to lower-case
       expect(await proxyBridge.hasRole(`0x${'0'.repeat(64)}`, defaultAdminSigner.address)).to.equal(true);
       expect(await proxyBridge.hasRole(WITHDRAW_ROLE, defaultAdminSigner.address)).to.equal(false);
       // Withdrawal by defaultAdminSigner should be rejected if it has not been granted WITHDRAW_ROLE
+      // sanity check on whether we need to make the defaultadminsigner.address to lower-case
```
cuongquangnam, line 84: hmm, the comment is not about this error, haha. It is more about my surprise that solidity error transforms the address to lowercase when emitting an error. So I check again whether I need to transform address into lowercase  when passing them into functions

weiyuan95 (approved):

PR #599, opened 2023-03-01
chore(ui-ux): improve ux on step 2 (dfc -> evm)

### LINKED ISSUE #562: Error code 3: when user send more than specified amount
```
Ver: 0.14.0
Network: TestNet
Flow: DFC → ETH 
Description: Error code 3: when user send more than specified amount
```

**Actual:**
– at **step 3 Validate**, if user encounters `Error code 3: BalanceNotMatched` 
– The balance in the wallet's address does not equal to the amount filled in the form
– user has no other option but to select x and restart the tx 

TBC

**To reproduce:**
Select Source Network: Ethereum
Input `0.02` dETH > select `Review transaction`
at **step 1 Initiate**, paste `Ethereum address` in text field > Continue
at **step 2 Transact**, use MLW to send `0.3` dETH > when tx is sent > go back to **Quantum**, select `Verify transfer`
at **step 3 Validate**, `error:` Error code 3: BalanceNotMatched > `note:` user get back 0.1 dETH as specified amount is 0.02 dETH

230224 Quantum - over sends amount

pierregee:
- as titled

Fixes #562

#### Developer Checklist:

- [x] Your UI implementation visually matched the rendered design\*
- [ ] Unit tests\*
- [ ] Added e2e tests\*
- [ ] Added translations\*


apps/web/src/components/commons/AlertInfoMessage.tsx
```diff
@@ -6,7 +6,7 @@ export default function AlertInfoMessage({
   containerStyle,
   textStyle,
 }: {
-  message: string;
+  message: string | JSX.Element;
```
fullstackninja864, line 9: can we avoid passing JSX.Element as props? we can use children property here

apps/web/src/styles/globals.scss
```diff
@@ -185,7 +185,6 @@ body {
   }
   .dark-bg-card-section {
     background: var(--ui-dark-card-section);
-    opacity: var(--tw-bg-opacity);
```
fullstackninja864, line 188: why its removed?

fullstackninja864 (changes requested):

apps/web/src/styles/globals.scss
pierregee: Yes, first image is with ` opacity: var(--tw-bg-opacity);` as you can see there's a diff on the colors (looks dim), if I don't remove it, the amount won't be prominent enough for the user to focus their eyes on
image
image

apps/web/src/components/commons/AlertInfoMessage.tsx
pierregee: Updated!

PR #634, opened 2023-03-02
fix(server): prevent creation of defichain private keys if it's not 2…

thedoublejay:
…4 words

The methods to generate a DeFiChain wallet accepts a string. This string can be anything even if it's not a valid 24 mnemonic words. This will prevent that to make sure that even if it's not set correctly on Env Variables, it would still work.

Fixes #

#### Developer Checklist:

- [ ] Unit tests\*
- [ ] Added e2e tests\*
- [ ] Added translations\*


lykalabrada (approved):

weiyuan95:
Let's use `validateMnemonicSentence` from jellyfish

lykalabrada (approved):

PR #710, opened 2023-03-08
chore(refactor): performance improvements

nattadex:
Moving the bridge status check api call to a `getServerSideProps()` function

Fixes #

#### Developer Checklist:

- [ ] Unit tests\*
- [ ] Added e2e tests\*
- [ ] Added translations\*


apps/web/src/pages/index.tsx
```diff
@@ -90,4 +92,20 @@ function Home() {
...
+  let isLoaded = false;
+  let isBridgeUp = true;
+  if (data) {
+    if (data?.isUp === true) {
+      isBridgeUp = true;
+    }
+    isLoaded = true;
+  }
+  return {
+    props: { isLoaded, isBridgeUp }, // will be passed to the page component as props
```
kevzzsk, line 107: `isBridgeUp` doesn't seem to be used anywhere. in `Home` component seems to only read `isLoaded`.

apps/web/src/pages/index.tsx
```diff
@@ -90,4 +92,20 @@ function Home() {
   );
 }
 
+export async function getServerSideProps() {
+  const res = await fetch(`https://wallet.defichain.com/api/v0/bridge/status`);
```
kevzzsk, line 96: Please add the error case and http error status case for non 200 cases

apps/web/src/pages/index.tsx
```diff
@@ -90,4 +90,18 @@ function Home() {
...
+  const data = await res.json();
+  let isBridgeUp = true;
+  if (data) {
+    if (data?.isUp === true) {
+      isBridgeUp = true;
```
kevzzsk, line 99: i think you have to flip the logic here, if not `isBridgeUp` will never be `false`.

you need to check `if (data?.isUp === false)` then set `isBridgeUp` to false

kevzzsk (changes requested):

apps/web/src/pages/index.tsx
nattadex: done!

apps/web/src/pages/index.tsx
nattadex: donee!

apps/web/src/pages/index.tsx
```diff
@@ -90,4 +90,29 @@ function Home() {
...
 
+export async function getServerSideProps() {
+  let isBridgeUp = true;
+
+  return fetch(`https://wallet.defichain.com/api/v0/bridge/status`)
+    .then((res) => Promise.all([res.json(), res.status]))
+    .then(([data, statusCode]) => {
+      if (statusCode === 200) {
+        if (data?.isUp === false) {
+          isBridgeUp = false;
```
thedoublejay, line 101:
```suggestion
          isBridgeUp = data?.isUp === false;
```

No need for the if statement anymore.

thedoublejay (changes requested):

thedoublejay (approved):

PR #737, opened 2023-03-10
feat(api): ethereum/stats API

### LINKED ISSUE #726: `stats` - `ethereum` - endpoint
Similar to https://github.com/BirthdayResearch/quantum/issues/725

We can use it an endpoint like `/ethereum/quantum/stats`. We only return daily transactions.

1. Number of total transactions
2. Number of confirmed transactions
3. Amounts bridged (total) per token

Need to cache it daily as well.

jordankzf:
- Add endpoint for ethereum statistics

Fixes #726

#### Developer Checklist:

- [x] Unit tests\*
- [ ] Added e2e tests\*
- [ ] Added translations\*


apps/server/src/types.ts
```diff
@@ -0,0 +1,2 @@
+export type ISO8601String = `${number}-${number}-${number}T${number}:${number}:${number}.${number}Z`;
+export type Erc20Token = "WBTC" | "USDT" | "USDC" | "ETH";
```
kevzzsk, line 2: there is already supported token list `SupportedEVMTokenSymbols`
https://github.com/BirthdayResearch/quantum/blob/8481ba96a959c66e14a08dd1887067ed74847165/apps/server/src/AppConfig.ts#L59-L64

apps/server/src/types.ts
```diff
@@ -0,0 +1,2 @@
+export type ISO8601String = `${number}-${number}-${number}T${number}:${number}:${number}.${number}Z`;
```
kevzzsk, line 1: I think we dont need the Time portion of the ISO8601. We can just expect a format of YYYY-MM-DD string

apps/server/src/ethereum/EthereumInterface.ts
```diff
@@ -0,0 +1,9 @@
+import { Erc20Token } from "src/types"
+
+export interface StatsModel {
+    totalTransactions: number,
+    confirmedTransactions: number,
+    amountBridged: {
+        [token in Erc20Token]?: number
```
kevzzsk, line 7: the value type should not be `number`, it should be `string` because we need to be precise with the DP

jordankzf:
To-do

- [x] Date filtering
- [x] Target Ethereum table
- [x] Unit tests
- [x] 6 Decimal places
- [x] Cache
- [ ] Manual testing (via real database entries)

apps/server/src/ethereum/services/EVMTransactionConfirmerService.ts
```diff
@@ -61,6 +63,81 @@ export class EVMTransactionConfirmerService {
...
+      amountBridged[token as SupportedEVMTokenSymbols] = amountBridgedBigN[token as SupportedEVMTokenSymbols]
+        .decimalPlaces(6, BigNumber.ROUND_FLOOR)
+        .toString();
+    });
+
+    return {
+      totalTransactions,
+      confirmedTransactions: confirmedTransactions.length,
+      amountBridged,
+      date: dateOnly,
```
kevzzsk, line 137: I don't think we need to return the `date` back in the response as the request will already have `date` params input from the caller

apps/server/src/ethereum/controllers/EthereumController.ts
```diff
@@ -2,18 +2,41 @@ import { Body, Controller, Get, Param, Post, UseGuards } from '@nestjs/common';
...
 
+  @Get('stats')
+  async getStats(@Param('date') date?: Iso8601String): Promise<StatsModel> {
+    return (await this.cache.get(
+      `ETH_STATS_${date}`,
+      async () => {
+        try {
+          return await this.evmTransactionConfirmerService.getStats(date);
+        } catch (e) {
+          throw new Error('API call for Ethereum statistics was unsuccessful', { cause: e });
```
kevzzsk, line 31: can you make sure that this throws a http error and return an error message to API ?

apps/server/src/ethereum/services/EVMTransactionConfirmerService.ts
```diff
@@ -61,6 +63,81 @@ export class EVMTransactionConfirmerService {
...
+        where: {
+          createdAt: {
+            gte: today.toISOString(),
+            lt: tomorrow.toISOString(),
+          },
+        },
+      }),
+
+      // Count only confirmed transactions
+      this.prisma.bridgeEventTransactions.findMany({
```
pierregee, line 87: Can we explore if we can move the token grouping + sum of each confirmed txn in db. A prisma equivalent of this to reduce looping 
```

SELECT bet."tokenSymbol", SUM(CAST (amount AS DECIMAL)) as "sumOfConfirmedTxn" 
FROM "BridgeEventTransactions" bet 
-- WHERE
group BY bet."tokenSymbol"
```

apps/server/src/ethereum/services/EVMTransactionConfirmerService.ts
```diff
@@ -61,6 +63,81 @@ export class EVMTransactionConfirmerService {
...
   }
 
+  async getStats(date?: Iso8601String): Promise<StatsModel> {
+    // Use today's date if no param is given
+    // NOTE: REMOVE THE TIMESTAMP UNLESS YOU WANT A 24-HOUR ROLLING WINDOW
+    const dateOnly = date ?? (new Date().toISOString().slice(0, 10) as Iso8601String);
+    const today = new Date(dateOnly);
+    today.setUTCHours(0, 0, 0, 0); // set to UTC +0
+    const tomorrow = new Date(today);
+    tomorrow.setDate(today.getDate() + 1);
```
pierregee, line 73:
```suggestion
    const tomorrow = new Date(today.getDate() + 1);
```

pierregee (changes requested):

apps/server/src/ethereum/services/EVMTransactionConfirmerService.ts
```diff
@@ -61,6 +63,81 @@ export class EVMTransactionConfirmerService {
...
+            gte: today.toISOString(),
+            lt: tomorrow.toISOString(),
+          },
+        },
+      }),
+
+      // Count only confirmed transactions
+      this.prisma.bridgeEventTransactions.findMany({
+        where: {
+          status: 'CONFIRMED',
```
pierregee, line 89: What is the expected return for`confirmedTransactions`? I think we should also check if `sendTransactionHash` column is not null which means that the user received the DFC equivalent of EVM token

apps/server/src/ethereum/services/EVMTransactionConfirmerService.ts
jordankzf: It's for confirmation reasons, especially when we are caching the result. In fact, I wish to alter it so that it returns the timestamp as well. This way, the caller will know how stale the data provided is, since it may be up to 23 hours 59 minutes 59 seconds old.

apps/server/src/ethereum/services/EVMTransactionConfirmerService.ts
jordankzf: This was my initial implementation.

Something like the following:

`await prisma.bridgeEventTransactions.groupBy({
    by: ['tokenSymbol'],
    _sum: {
      amount: true,
    },
  });`
  
This issue here is that precision is lost due to the way the amounts are summed up internally. Prisma cannot do `BigNumber` operations.

apps/server/src/ethereum/services/EVMTransactionConfirmerService.ts
jordankzf: `confirmedTransactions` is the total record/row count of transactions that are successful.

`sendTransactionHash` non-null check has been added.

apps/server/src/ethereum/services/EVMTransactionConfirmerService.ts
jordankzf: Either way, I think it's still good practice to return the queried date, so that the user can verify that the date param was parsed correctly, since we accept any string arbitrarily.

Let me know if you prefer to have it removed.

apps/server/src/ethereum/services/EVMTransactionConfirmerService.ts
jordankzf: This suggested change overflows the tomorrow const to `1970-01-01T00:00:00.010Z`

It does not seem equivalent.

jordankzf:
API Definition Documentation updated here: https://docs.google.com/document/d/1W3Vnjd57_D2U24ivBdKihaQRLOogD3RKv4zHovFqCAM/edit

apps/server/src/types.ts
```diff
@@ -0,0 +1,2 @@
+export type Iso8601DateOnlyString = `${number}-${number}-${number}`;
```
thedoublejay, line 1: We shouldn't have this common types. It's a bad practice that might be re-used by other devs and be filled up with mixed data. I suggest we keep it where it's only needed. (Stats module).

apps/server/test-i9n/testing/Statistics.i9n.ts
```diff
@@ -0,0 +1,88 @@
+import { PostgreSqlContainer, StartedPostgreSqlContainer } from '@birthdayresearch/sticky-testcontainers';
+
```
thedoublejay, line 2: Same as this, let's move this to a ethereum specific file. If we support more chains in the future, we can have "bitcoin/stats, polygon/stats" etc.

apps/server/src/ethereum/EthereumModule.ts
```diff
@@ -19,9 +20,10 @@ import { EVMTransactionConfirmerService } from './services/EVMTransactionConfirm
     WhaleApiClientProvider,
     DeFiChainTransactionService,
     EVMTransactionConfirmerService,
+    SemaphoreCache,
   ],
   controllers: [EthereumController],
-  imports: [EthersModule],
+  imports: [EthersModule, CacheModule.register({ max: 10_000 })],
```
thedoublejay, line 26: Would this affect other modules under Ethereum? Can we only cache the specifc service/module?

apps/server/src/ethereum/controllers/EthereumController.ts
```diff
@@ -1,19 +1,51 @@
...
+            {
+              cause: e,
+            },
+          );
+        }
+      },
+      {
+        ttl: 3600_000 * 24, // 1 day
+      },
+    )) as StatsModel;
```
thedoublejay, line 46: Why do we need to cache it? Maybe better to add it on the cache itself? Like this.cache.get<StatsModel>()

thedoublejay (changes requested):

apps/server/src/ethereum/controllers/EthereumController.ts
```diff
@@ -1,19 +1,51 @@
...
+    protected readonly cache: SemaphoreCache,
+  ) {}
 
   @Get('balance/:tokenSymbol')
   async getBalance(@Param('tokenSymbol') tokenSymbol: SupportedEVMTokenSymbols): Promise<string> {
     return this.evmTransactionConfirmerService.getBalance(tokenSymbol);
   }
 
+  @Get('stats/')
+  async getStats(@Query('date') date?: Iso8601DateOnlyString): Promise<StatsModel> {
```
thedoublejay, line 24: Does this work as DTO? https://docs.nestjs.com/techniques/validation

thedoublejay (changes requested):

apps/server/src/ethereum/services/EVMTransactionConfirmerService.ts
kevzzsk: As you mentioned, stale data will only happen on today's date and the purpose of the endpoint is not to provide realtime update but more to allow historical query of data. So I wouldn't be too worried by the stale data for today's query given it is not yet completed.

secondly, about verifying whether the date param was parsed correctly, we should always reject param that does not conform to the specified format YYYY-MM-DD, so user will know always know that the data is always correct and wrong query will not have data.

jordankzf: 1. So you want to remove both `date` and `cacheTime` from the return?
2. This is the current implementation. If you provide the `&date=` without anything, it will be rejected.

apps/server/src/ethereum/EthereumModule.ts
jordankzf: It should not. `this.cache.get()` must be explicitly called in order to cache the return.

apps/server/src/ethereum/controllers/EthereumController.ts
jordankzf: Yes, it's cached on line 25 via `this.cache.get()`

apps/server/src/ethereum/services/EVMTransactionConfirmerService.ts
kevzzsk: Yeah, I think `date` and `cacheTime` will not be meaningful for end users

apps/server/src/ethereum/controllers/EthereumController.ts
thedoublejay: Sorry my message was incorrect. My question was why do we need to "typecast" it as StatsModel. Can we do something like cache.get<StatsModel>()

apps/server/test-i9n/testing/Statistics.i9n.ts
jordankzf: It's the same pattern used in `VersionController.ts` and `WhaleWalletController.ts`. Let me see if it's possible to avoid typecasting.

apps/server/src/ethereum/services/EVMTransactionConfirmerService.ts
jordankzf: Removed.

apps/server/test-i9n/testing/Statistics.i9n.ts
jordankzf: The reason it was typecasted is because `this.cache.get()` can possibly return `undefined`. The return type at its source is defined by `async-mutex`. For this reason, it cannot be edited unless overwritten.

I made this change instead to avoid typecasting:
`async getStats(@Query('date') date?: Iso8601DateOnlyString): Promise<StatsModel | undefined> {`

apps/server/src/ethereum/controllers/EthereumController.ts
jordankzf: The reason it was typecasted is because this.cache.get() can possibly return undefined. The return type at its source is defined by async-mutex. For this reason, it cannot be edited unless overwritten.

I made this change instead to avoid typecasting:
async getStats(@Query('date') date?: Iso8601DateOnlyString): Promise<StatsModel | undefined> {

apps/server/src/ethereum/services/EVMTransactionConfirmerService.ts
```diff
@@ -61,6 +63,84 @@ export class EVMTransactionConfirmerService {
     return ethers.utils.formatUnits(balance, assetDecimalPlaces);
   }
 
+  async getStats(date?: Iso8601DateOnlyString): Promise<StatsModel> {
```
kevzzsk, line 66: can we add a simple sanity check on the date passed in? like rejecting when the date is in the future (more than todays date). This is to prevent useless query reaching our DB

jordankzf: Added

apps/server/src/ethereum/controllers/EthereumController.ts
jordankzf: Converted to DTO

apps/server/src/ethereum/controllers/EthereumController.ts
```diff
@@ -1,19 +1,51 @@
...
   }
 
+  @Get('stats/')
+  async getStats(@Query('date') date?: StatsQueryDto): Promise<StatsDto | undefined> {
+    return this.cache.get(
+      `ETH_STATS_${date ?? 'TODAY'}`,
+      async () => {
+        try {
+          const statsModel: StatsModel = await this.evmTransactionConfirmerService.getStats(date);
+          return new StatsDto(statsModel);
```
jordankzf, line 29: I _think_ this is how DTO validation is done, not entirely sure.

CC: @thedoublejay @nattadex

thedoublejay: Looks ok!

thedoublejay: Why do we need to have a StatsDto? It looks like it's the same as StatsModel?

thedoublejay (changes requested):

apps/server/.eslintignore
```diff
@@ -0,0 +1 @@
+src/types.ts
```
kevzzsk, line 1: this should not be needed anymore right

apps/server/src/utils/StatsUtils.ts
```diff
@@ -0,0 +1,2 @@
+export type Iso8601DateOnlyString = `${number}-${number}-${number}`;
+export type Iso8601String = `${number}-${number}-${number}T${number}:${number}:${number}.${number}Z`;
```
kevzzsk, line 2: this should also not be used anymore

apps/server/test-i9n/testing/ethereum/Statistics.i9n.ts
kevzzsk, line 1: Could you also add the case where user input a future date?

kevzzsk (changes requested):

apps/server/.eslintignore
jordankzf: removed

apps/server/src/utils/StatsUtils.ts
jordankzf: removed

apps/server/test-i9n/testing/ethereum/Statistics.i9n.ts
jordankzf: added

apps/server/src/ethereum/controllers/EthereumController.ts
jordankzf: Removed!

apps/server/src/ethereum/services/EVMTransactionConfirmerService.ts
```diff
@@ -61,6 +62,88 @@ export class EVMTransactionConfirmerService {
     return ethers.utils.formatUnits(balance, assetDecimalPlaces);
   }
 
+  async getStats(date?: StatsQueryDto): Promise<StatsDto> {
+    // Use today's date if no param is given
+    // NOTE: REMOVE THE TIMESTAMP UNLESS YOU WANT A 24-HOUR ROLLING WINDOW
+    const dateOnly = date ?? new Date().toISOString().slice(0, 10);
+    const dateFrom = new Date(dateOnly as string);
```
kevzzsk, line 69: we also need to take care of `date` is a valid date string? because if we pass like `2023-00-00` it will pass the format check but it is not a valid date. It will probably cause server error.

maybe add this to the test too

apps/server/src/ethereum/controllers/EthereumController.ts
```diff
@@ -1,19 +1,54 @@
...
     return this.evmTransactionConfirmerService.getBalance(tokenSymbol);
   }
 
+  @Get('stats/')
+  async getStats(@Query('date') date?: StatsQueryDto): Promise<StatsDto | undefined> {
+    return this.cache.get(
+      `ETH_STATS_${date ?? 'TODAY'}`,
+      async () => {
+        try {
+          // const statsModel = await this.evmTransactionConfirmerService.getStats(date);
```
kevzzsk, line 28: we can remove not needed comments too

kevzzsk (changes requested):

apps/server/src/ethereum/services/EVMTransactionConfirmerService.ts
jordankzf: `ethereum/stats?date=2023-00-00` returns `{"status":500,"error":"API call for Ethereum statistics was unsuccessful: Invalid time value"}`

JS's native `Date` object constructor will fail if `date` is not valid, and the `try...catch` statement which wraps the `getCache()` method gracefully handles this exception.

I used this design pattern so that the error will not be cached. The resulting payload will only be cached if it was computed successfully.

apps/server/src/ethereum/controllers/EthereumController.ts
jordankzf: removed

kevzzsk (approved):

thedoublejay (approved):

PR #745, opened 2023-03-13
fix(core): only limit all amounts to 6 decimal places maximum

lykalabrada:
Claim with small amount is failing for tokens with 6 decimal places (USDC, EUROC). This happens when fee is deducted from the amount and the resulting difference has more than six decimals.

Fixes #

#### Developer Checklist:

- [ ] Unit tests\*
- [ ] Added e2e tests\*
- [ ] Added translations\*


pierregee (changes requested):
image

pierregee (approved):

thedoublejay (approved):

